PR #49, opened 2026-06-18
Fill missing home hero logos from TMDB artwork

wifi-x-smasher:
## What changed

This lets home catalog hero items fill a missing logo from TMDB artwork when TMDB artwork enrichment is enabled.

## Why

Some catalog preview items have a backdrop and title, but no logo. On desktop that leaves the home hero falling back to plain title text even though TMDB has a localized logo available. This keeps add-on supplied logos untouched and only fills the missing case.

## Scope

- Home catalog hero items only.
- Uses the existing TMDB artwork setting and API key checks.
- Fetches only TMDB image metadata for the logo.
- Does not change details pages, player behavior, or hero metadata text.

## Testing

- Ran `./gradlew.bat :composeApp:desktopTest --tests "com.nuvio.app.features.tmdb.TmdbMetadataServiceTest" --tests "com.nuvio.app.features.home.HomeCatalogParserTest" --project-prop=nuvio.webview2.dir=S:/Nuvio/NuvioDesktop/.tmp/nuget/microsoft.web.webview2/1.0.3351.48 --no-daemon --stacktrace`
- Manually checked the desktop home hero locally with an item that previously showed plain title text; it now shows logo artwork when TMDB provides one.

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/features/home/HomeRepository.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/features/home/HomeRepository.kt
@@ -11,6 +11,8 @@ import com.nuvio.app.features.collection.CollectionSource
 import com.nuvio.app.features.collection.TmdbCollectionSourceResolver
 import com.nuvio.app.features.collection.catalogRouteKey
 import com.nuvio.app.features.collection.findCollectionCatalog
+import com.nuvio.app.features.tmdb.TmdbMetadataService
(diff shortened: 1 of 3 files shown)


wifi-x-smasher:
Updated the PR to address the Copilot review comments.

The follow-up commit scopes the home hero logo cache by TMDB language, avoids applying cached TMDB logos when artwork enrichment is disabled, and guards the preview-logo cache with a mutex so concurrent enrichment does not read/write the map unsafely.

Issue #33, opened 2026-06-17
[Feature]: Please let us use Arrows keys to navigate


mosestyle:
### Area (tag)

Keyboard / Mouse input

### Problem statement

Currently, NuvioDesktop relies heavily on mouse navigation. When browsing movies, TV shows, search results, seasons, episodes, or source lists, users must constantly move between the keyboard and mouse.

This can make navigation slower and less convenient, especially for users who prefer keyboard controls or use NuvioDesktop on a TV, couch setup, or media center PC.

I would like to be able to navigate the interface using the keyboard arrow keys.

### Proposed solution

Add keyboard navigation support throughout NuvioDesktop.

Suggested controls:

- Keyboard arrows keys : Move between items
- Enter : Open/select highlighted item
- Backspace or Esc : Go back to the previous screen
- Page Up / Page Down : Faster scrolling (optional)

A visible highlight/focus indicator would show which item is currently selected.

This would make NuvioDesktop much easier to use without constantly reaching for the mouse and would provide a more TV-friendly and accessibility-friendly experience.

### Are you planning to implement this yourself?

No, this is only a proposal

ReaLx3m:
Yup, id like to see this one too, this a must for Desktop imo. Will also make possible to use it with a remote on Windows based HTPCs.

Rejoyceslav:
I have a LG MagicRemote mapped to Windows's input and am using the PC in "TV mode". Would also be nice to enable scroll to control the volume

stablestud:
+1 from me, I wonder if NuvioWeb GUI can be brought into the Desktop App, then we have a completely nice setup for HTPC PCs using controllable with controller or simply bluetooth remotes

stablestud:
This is extended by https://github.com/NuvioMedia/NuvioDesktop/issues/305

PR #180, opened 2026-06-29
Keyboard Arrow Navigation

aelrased:
## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [x] Small maintenance only, with no UI or behavior change
- [ ] Docs accuracy fix

## UI / behavior impact

- [x] No UI change
- [x] No behavior change

## Policy check

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/ShelfComponents.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/ShelfComponents.kt
@@ -270,6 +270,7 @@ fun NuvioPosterCard(
                     shape = cardShape,
                     surface = NuvioCardDepthSurface.Posters,
                 )
+                .nuvioFocusBorder(cardShape)
                 .posterCardClickable(
                     onClick = onClick,
                     onLongClick = onLongClick,
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ scripts
 
 # Local MPVKit iOS build environment (sparse APFS image, see MPVKit docs)
 .mpvkit-build.sparseimage
+
+composeApp/hs_err_pid*.log
+hs_err_pid*.log
--- a/composeApp/src/androidMain/kotlin/com/nuvio/app/core/ui/PlatformKeyboardNavigation.android.kt
+++ b/composeApp/src/androidMain/kotlin/com/nuvio/app/core/ui/PlatformKeyboardNavigation.android.kt
@@ -1,0 +1,8 @@
+package com.nuvio.app.core.ui
+
+import androidx.compose.runtime.Composable
+
+@Composable
+actual fun PlatformKeyboardNavigation() {
+    // No-op on Android — keyboard navigation is handled by the platform.
+}
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/App.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/App.kt
@@ -16,6 +16,7 @@ import coil3.request.crossfade
 import coil3.svg.SvgDecoder
 import com.nuvio.app.core.ui.NativeProfileSwitcherController
 import com.nuvio.app.core.ui.NuvioTheme
+import com.nuvio.app.core.ui.PlatformKeyboardNavigation
 import com.nuvio.app.core.ui.configurePlatformImageLoader
 import com.nuvio.app.core.ui.desktopUiScaleForWindow
 import com.nuvio.app.features.settings.ThemeSettingsRepository
@@ -102,6 +103,7 @@ internal fun AppEnvironment(content: @Composable () -> Unit) {
             amoled = amoledEnabled,
             desktopUiScale = desktopUiScaleForWindow(maxWidth.value, maxHeight.value),
         ) {
+            PlatformKeyboardNavigation()
             content()
         }
     }
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/NavigationBar.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/NavigationBar.kt
@@ -266,11 +266,13 @@ private class NuvioNavigationBarScopeImpl(
         )
 
         with(rowScope) {
+            val navItemShape = RoundedCornerShape(NuvioTokens.Radius.full)
             Column(
                 modifier = modifier
                     .weight(1f)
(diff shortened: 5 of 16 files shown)


aelrased:
## Summary — Keyboard Arrow Navigation

### Problem
- `onPreviewKeyEvent` does not work without a focused element in Compose Desktop
- `LazyRow`/`LazyColumn` consume arrow key events before they reach the top level

### Solution
Use AWT `KeyEventDispatcher` (same approach as `installDesktopAppFullscreenShortcuts`) to intercept arrow keys outside the Compose Focus system entirely.

### New Files (5)
| File | Description |
|------|-------------|
| `PlatformKeyboardNavigation.kt` | expect declaration |
| `PlatformKeyboardNavigation.desktop.kt` | AWT KeyEventDispatcher → focusManager.moveFocus() + fallback FocusDirection.Next |
| `PlatformKeyboardNavigation.android.kt` | no-op |
| `PlatformKeyboardNavigation.ios.kt` | no-op |
| `NuvioFocusable.kt` | Modifier.nuvioFocusBorder(shape) — draws a 3dp border using the theme's `focusRing` color around the element's actual shape |

### Modified Files (5)
| File | Change |
|------|--------|
| `App.kt` | Moved `PlatformKeyboardNavigation()` to `App()` root (works on all screens). Added `nuvioFocusBorder` to `TabletTopPillItem` and `DesktopSidebarItem` |
| `NavigationBar.kt` | Added `nuvioFocusBorder` to all 3 `NavItem` overloads |
| `ProfileSelectionScreen.kt` | Added `nuvioFocusBorder` to `ProfileAvatarCard` and `AddProfileCard` |
| `ShelfComponents.kt` | Added `nuvioFocusBorder` to `NuvioPosterCard` |
| `SettingsComponents.kt` | Added `nuvioFocusBorder` to `SettingsSidebarItem`, `SettingsNavigationRow`, `SettingsSwitchRow` |

### Focusable Elements
Bottom nav bar, desktop sidebar, tablet top pills, poster cards, profile avatar cards, add profile button, and settings items.

composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/ShelfComponents.kt
```diff
@@ -235,6 +235,7 @@ fun NuvioPosterCard(
     Column(
         modifier = Modifier
             .posterCardClickable(onClick = onClick, onLongClick = onLongClick)
+            .nuvioFocusBorder(cardShape)
```
DustinBryant: This part isn't working for the poster cards for me because .posterCardClickable is swallowing the events it needs. If you put .nuvioFocusBorder first ahead of .posterCardClickable you get the desired functionality.

aelrased:
@DustinBryant Good catch! I've pushed a fix — moved .nuvioFocusBorder(cardShape) before .posterCardClickable(...) so the focus ring can receive keyboard events before the clickable modifier intercepts them. The change is on the keyboard-nav branch now.

DustinBryant:
@KhooLy can we get some eyes on this?

It would address:
#267 
#33 
#305 

This change would bring in many more users with Home Theatre PCs (HTPC). Right now we're mostly stuck with Stremio but desperately want Nuvio.

KhooLy:
Good direction, but the global KeyEventDispatcher currently consumes every arrow key press across the app, even when focus movement isn't handled. This can interfere with text fields, sliders, scrolling and other components that need their own arrow-key behavior. Please only consume the event when appropriate, and make the focus/navigation behavior desktop-only rather than changing shared mobile focus behavior. Also please rebase onto current Dev and update the PR description/testing, since this is a behavior/UI feature rather than maintenance.

composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/ShelfComponents.kt
aelrased: Thanks for catching that! Fixed in the latest push.

**What changed:**
- Moved .nuvioFocusBorder() from the inner Box to the Column modifier on SeasonPosterButton, placed before .posterCardClickable so the focus ring receives keyboard events before the click handler swallows them.
- The episode card already had the correct order (.nuvioFocusBorder before .posterCardClickable), so no change needed there.

All other components (StreamCard, StreamsScreen filter chips, DetailActionButtons, DetailCastSection avatars) were already correct.

aelrased:
Thanks for the review feedback - all three points have been addressed in the latest push:

1. **Smart arrow key consumption** - PlatformKeyboardNavigation.desktop.kt now inspects the focused AWT component before consuming arrow keys. It only consumes when the focus is on a Compose-managed focusable (not JTextField, JTextArea, JSpinner, JSlider, JScrollPane, or Scrollable), so text fields, sliders, scroll panes and other AWT components retain their native arrow-key behavior.

2. **Desktop-only** - .focusable() in NuvioFocusable.kt is now guarded with isDesktop, so mobile platforms are completely unaffected.

3. **Rebased onto current Dev** - Branch has been rebased onto upstream/Dev (21aabeeb).

Other changes in this push:
- Fixed modifier ordering on SeasonPosterButton - .nuvioFocusBorder() moved before .posterCardClickable() per the inline review comment, so the focus ring receives keyboard events before the click handler can swallow them.
- Added focus rings to StreamCard, StreamsScreen filter chips, DetailActionButtons (Play / More), DetailCastSection avatars, and all episode card variants.

Build verified successfully. Ready for another review.

DustinBryant:
Sorry, just now getting around to testing your changes.

1. Moving vertically with arrow keys is a little off here. I have to hit the same direction twice just for it to highlight and move to the catalog item above or below. This is because the first arrow press will hit the catalog wrapper first before getting to the child card. I think it's more intuitive to go directly to the card and if not to at least put on a focus ring so we know where we currently are.
2. I have a Continue Watching section with larger cards that I can't get the keyboard to focus on any of them.
3. Though we can get to the side menu via arrows, there's no indication to what is focused.
4. Enter on a card should take you to the details of that card. Right now the enter key does nothing on a highlighted card.
5. Within the details of a card I can't seem to get any keyboard interaction. I should be able to move to the play button and trailer section with keys.
6. Esc key should take you back from any screen you are on including while watching content. If I'm in the detailed view if I hit esc it should take me back to my catalogs.
7. The side menu settings, library, and search I'm having trouble getting focus on all the parts of all. Maybe for those screens you start focus on the first item in the screen with focus capabilities. For instance, if I go to settings then auto focus on Account. This may actually help other screens where focus isn't getting captured correctly.

That's all for now.

Issue #78, opened 2026-06-20
[Bug]: Network issues after some time addons stream and metadata requests intermittently fail with "Can't assign requested address"- macOS


amackarrey:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.1.6-alpha

### Install method

GitHub Release desktop installer/package

### Platform

macOS (desktop)

### Desktop hardware

MacBook Air m4

### OS version

macOS 26.5.1

### Area (tag)

Streams / Sources (selection, loading, errors)

### Bug description

After using Nuvio for a short time, addon requests start failing intermittently. The first stream or metadata request can work normally, but later requests sometimes fail or return no streams until I retry or restart the app.

This affects AIOStreams and metadata providers. In my case, AIOStreams returned streams successfully once, then a later stream fetch failed. Metadata requests also failed around the same time.

The terminal log shows this error:

java.io.IOException: Can't assign requested address

The same error appears for:
- AIOStreams stream fetching
- AIOMetadata metadata fetching
- Bingecat metadata fetching
- MDBList metadata enrichment
it whole app go puff for sometime in short. i noticed it also happens in windows per user query

Restarting Nuvio fixes the issue temporarily, but the problem comes back.

### Steps to reproduce

Open Nuvio on macOS.
2. Open a show or movie.
3. Fetch streams using AIOStreams.
4. Navigate between items or episodes and fetch streams/metadata again.
5. After one or more requests, some addon requests fail or show no streams.
6. Restart Nuvio.
7. Fetching works again temporarily.

### Expected behavior

Nuvio should continue fetching streams and metadata reliably without needing an app restart.

### Actual behavior

ddon requests intermittently fail. Streams may show no results, metadata may fail to load, and retrying is hit or miss. Restarting the app temporarily fixes it.

### Frequency

Always

### Did this work before?

No, it never worked

### Logs (required for crash reports)

Relevant log excerpt, redacted:

Debug: (StreamsRepo) Found 1 addons for stream type=series id=...
Debug: (StreamsRepo) Fetching streams from: [AIOSTREAMS_URL_REDACTED]
Debug: (StreamsRepo) Got 15 streams from AIOStreams-N
[...]

PR #185, opened 2026-06-29
Fix plugin provider freeze caused by blocking fetch on the QuickJS runtime thread

fmustafayaman:
## Summary

Fixes a freeze where JS **plugin** scrapers stop returning streams after a while and only recover after a full app restart. The streams screen gets stuck in the loading state and no plugin ever completes, even though the same providers worked moments earlier.

The root cause is in the plugin runtime's fetch bridge: `__native_fetch` was a synchronous binding that wrapped the suspending `httpRequestRaw` in `runBlocking`, so every plugin network request blocked a `Dispatchers.Default` thread for the whole request and ignored cancellation. Under concurrency (or a stalled endpoint) the bounded Default pool gets fully blocked and the runtime can no longer make progress.

This PR binds `__native_fetch` as an async (suspend) binding and calls `httpRequestRaw` directly (no `runBlocking`), and `await`s it from the JS `fetch` polyfill.

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

`JsRuntime` runs QuickJS on the shared `Dispatchers.Default` dispatcher:

```kotlin
// JsRuntime.kt
quickJs(dispatcher) { block() } // dispatcher = Dispatchers.Default
```

`FetchBridge` then registered the network entry point as a **synchronous** function that blocks on the suspending HTTP call:

```kotlin
// before
runtime.function("__native_fetch") { args ->
    ...
    performNativeFetch(...) // val response = runBlocking { httpRequestRaw(...) }
}
```

Consequences:

- Each plugin `fetch()` parks one `Dispatchers.Default` thread for the entire request duration.
- `Dispatchers.Default` is bounded by the CPU core count, so a handful of concurrent or slow plugin requests can block every thread in the pool at once.
- `runBlocking` does not observe coroutine cancellation, so neither the per-plugin `withTimeout(PLUGIN_TIMEOUT_MS)` in `PluginRuntime` nor `activeJob.cancel()` in `StreamsRepository` can interrupt an in-flight request.

Once the pool is starved, no other coroutine scheduled on `Dispatchers.Default` can run and the plugin runtime appears frozen until the process is restarted. This is easy to hit when several plugins run together (e.g. moving to the next episode a few times) or when a provider talks to a slow/cold endpoint.

### Old vs new behavior

- **Old:** a plugin network call blocks a Default-pool thread until it returns; enough concurrent/slow calls starve the dispatcher and freeze all plugins until restart.
- **Broken/unwanted:** plugins stop returning streams; timeouts and cancellation have no effect; only a restart recovers.
- **New:** the network call suspends instead of holding a thread, so the dispatcher keeps making progress, and `withTimeout` / cancellation can actually interrupt a stuck request.

## The fix

`composeApp/src/fullCommonMain/kotlin/.../network/FetchBridge.kt`
- Register `__native_fetch` with `asyncFunction` instead of `function`.
- Make `performNativeFetch` a `suspend` function and call `httpRequestRaw(...)` directly instead of through `runBlocking`.
- Rethrow `CancellationException` so cancellation is not swallowed by the existing error-mapping `catch`.

`composeApp/src/fullCommonMain/kotlin/.../js/JsBindings.kt`
- `await __native_fetch(...)` in the `fetch` polyfill, since the binding now returns a promise. The polyfill function is already `async`, so this is a one-line change.

No public API, dependency, or architecture change. `quickjs-kt` already supports async bindings (`asyncFunction`), so no new dependency is introduced.

## Desktop scope

The changed files are in the plugin runtime (`fullCommonMain`) used by the desktop app. This is a desktop stability fix for installed JS scrapers.

## Issue or approval

Related to #78 (same user-visible symptom: provider requests stop working until the app is restarted). That report and #104 target the addon HTTP path (socket/port exhaustion from per-request `HttpClient`); this PR fixes a **separate** root cause in the plugin runtime (Default-pool starvation from blocking fetch), so the two are complementary and do not overlap in code.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Testing

- Built and ran the desktop app on macOS (Apple Silicon).
- Before: after a few rounds of concurrent plugin fetches (switching episodes), plugins stopped returning streams and the screen stayed in the loading state until restart.
- After: plugins keep returning streams; a single slow/stalled request times out on its own without taking down the rest of the runtime, and switching episodes repeatedly no longer freezes plugin fetching.

## Policy check

- [x] This PR is scoped to the desktop app and shared code required for desktop behavior.
- [x] Behavior change fixes a documented bug/regression.
- [x] This PR does not bundle unrelated refactors, cleanups, or formatting.
- [x] This PR does not add dependencies, architecture changes, or migrations.
- [x] I listed the testing performed above.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/composeApp/src/fullCommonMain/kotlin/com/nuvio/app/features/plugins/runtime/js/JsBindings.kt
+++ b/composeApp/src/fullCommonMain/kotlin/com/nuvio/app/features/plugins/runtime/js/JsBindings.kt
@@ -47,7 +47,7 @@ internal object JsBindings {
             var headers = __normalize_fetch_headers(options.headers);
             var body = options.body || '';
             var followRedirects = options.redirect !== 'manual';
-            var result = __native_fetch(url, method, JSON.stringify(headers), body, followRedirects);
+            var result = await __native_fetch(url, method, JSON.stringify(headers), body, followRedirects);
             var parsed = JSON.parse(result);
             return {
                 ok: parsed.ok,
--- a/composeApp/src/fullCommonMain/kotlin/com/nuvio/app/features/plugins/runtime/network/FetchBridge.kt
+++ b/composeApp/src/fullCommonMain/kotlin/com/nuvio/app/features/plugins/runtime/network/FetchBridge.kt
@@ -2,10 +2,10 @@ package com.nuvio.app.features.plugins.runtime.network
 
 import co.touchlab.kermit.Logger
 import com.dokar.quickjs.QuickJs
-import com.dokar.quickjs.binding.function
+import com.dokar.quickjs.binding.asyncFunction
 import com.nuvio.app.features.addons.httpRequestRaw
 import com.nuvio.app.features.plugins.runtime.host.HostModule
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.CancellationException
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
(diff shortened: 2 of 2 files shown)


fmustafayaman:
Ekran Resmi 2026-06-29 18 43 56

PR solves this issue. After watching an episode from plugin provider it wont fetch next one. with this patch its just works as intended

Issue #16, opened 2026-06-16
[Bug]: Screen Flickering on Windows


Tether1776:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.1.1-alpha

### Install method

GitHub Release desktop installer/package

### Platform

Windows (desktop)

### Desktop hardware

Windows 11: 5800x3D/RTX 5080

### OS version

Windows 11 24H2

### Area (tag)

UI / Layout

### Bug description

The application UI looks like it is flickering, I think it is related to G-sync.

### Steps to reproduce

Open application

Use UI normally

View User Interface Flickering/Strobing

### Expected behavior

No UI flickering

### Actual behavior

The screen flickers during usage
There are no errors or logs

### Frequency

Always

### Did this work before?

Not sure

### Logs (required for crash reports)

```shell

```

Warkratos:
Hello, use NVIDIA Profile Inspector, and import this profile,  save the code as nuvio.nip, start nvidia profile inspector and import it.

```
<?xml version="1.0" encoding="utf-16"?>
<ArrayOfProfile>
  <Profile>
    <ProfileName>Nuvio Media</ProfileName>
    <Executeables>
      <string>nuvio.exe</string>
    </Executeables>
    <Settings>
      <ProfileSetting>
        <SettingNameInfo />
        <SettingID>4565791</SettingID>
        <SettingValue>0</SettingValue>
        <ValueType>Dword</ValueType>
      </ProfileSetting>
      <ProfileSetting>
        <SettingNameInfo>NVIDIA Predefined Ansel Usage</SettingNameInfo>
        <SettingID>271965065</SettingID>
        <SettingValue>0</SettingValue>
        <ValueType>Dword</ValueType>
      </ProfileSetting>
      <ProfileSetting>
        <SettingNameInfo>Do not display this profile in the Control Panel</SettingNameInfo>
        <SettingID>275602687</SettingID>
        <SettingValue>0</SettingValue>
        <ValueType>Dword</ValueType>
      </ProfileSetting>
      <ProfileSetting>
        <SettingNameInfo />
        <SettingID>276275482</SettingID>
        <SettingValue>0</SettingValue>
        <ValueType>Dword</ValueType>
[...]

PR #275, opened 2026-07-21
fix(desktop): add a renderer setting to fix UI flicker on high refresh rate displays

### LINKED ISSUE #269: [Bug]: UI flicker and low frame pacing on high refresh rate Windows displays
### Pre-flight checks

### App version / commit

0.1.13-alpha

### Install method

Built from source

### Platform

Windows

### OS version

Windows 11

### Desktop hardware

Windows 11 desktop, 360Hz display

### Bug description

On a high refresh rate display the interface flickers and feels like it is running at a much lower frame rate than the monitor. This affects the Compose interface (scrolling, animations, navigation). Video playback itself is unaffected, since that renders through the native mpv pipeline rather than Skiko.

Related: #16 was closed, but the problem still reproduces on high refresh rate monitors.

### Steps to reproduce

1. Run the app on a high refresh rate display (reproduced at 360Hz).
2. Scroll the home screen or move between screens.

### Expected behavior

The interface renders smoothly at the display refresh rate.

### Actual behavior

Visible flicker and a low frame-pace feel, roughly like 30fps, while video playback stays smooth.

### Additional context

Skiko defaults to Direct3D on Windows, and its present appears to misalign on high refresh rate displays. Forcing Skiko to OpenGL (`skiko.renderApi=OPENGL`) removes the flicker entirely on the same hardware, which is why this looks like a renderer issue rather than a Compose one.

OpenGL is not the right default for everyone, so the useful fix is likely to keep Direct3D as the default and expose the renderer as a setting for affected users.

Laskco:
## Summary

Adds an off-by-default Windows setting that switches the interface from Direct3D to OpenGL, which fixes flicker and choppiness on high refresh rate monitors. Nothing changes unless you turn it on.

## PR type

- [ ] Reproducible bug fix
- [x] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

On a high refresh rate monitor the interface flickers and feels like it is running at roughly 30fps, no matter how fast the display actually is. Scrolling and screen transitions are the most obvious. Video playback is fine, because video is rendered by mpv, not by the interface renderer (#269, and #16 which was closed but still reproduces).

The cause is the renderer. Compose draws the interface through Skiko, which uses Direct3D by default on Windows, and its frame presentation does not line up properly on these displays. Switching Skiko to OpenGL removes the flicker completely on the same machine, with no other changes.

Since this depends on the specific monitor and GPU, OpenGL is not a safe default for everyone. So instead of changing the default, this exposes it as a setting that affected users can turn on.

## Desktop scope

Windows desktop. The setting only appears on Windows.

Most of the changed files are the standard plumbing for showing a desktop-only setting in the shared settings screen: the desktop implementation plus `Platform` expect/actual, with the Android and iOS versions being no-ops.

## Issue or approval

Fixes #269. Related to #16, which was closed but still reproduces on high refresh rate monitors.

Flagging for maintainer judgement: this does add a visible setting. If that counts as a directional change rather than a bug fix, say so and I will close it. It is written to be completely inert unless enabled.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change
- [x] UI changed only to fix a documented glitch/bug

## Policy check

## Scope boundaries

Three deliberate choices:

- **Direct3D stays the default.** Anyone who does not enable the setting sees no change whatsoever.
- **The preference is stored per machine**, in its own store, so it never syncs between devices. The right renderer depends on the monitor you are sitting in front of, so syncing it would be actively wrong.
- **An existing `skiko.renderApi` property or environment variable still wins**, so anyone already pinning a renderer is unaffected.

No renderer is forced, no other setting is touched, and it requires a restart to apply (stated in the setting description).

## Testing

Windows 11, 2560x1440 at 360Hz, built from source, 0.1.13-alpha.

- Setting off (the default): runs on Direct3D exactly as before.
- Setting on, after restart: the flicker and choppiness are gone on the same hardware.
- The preference persists across restarts and is read before the interface starts.
- Video playback and NVIDIA RTX Super Resolution are unaffected either way, since mpv renders through its own D3D11 pipeline independently of the interface renderer.

Not tested on macOS or Linux; the setting is Windows-only.

## Screenshots / Video

A screenshot cannot show this. The bug is frame timing, so it does not survive a still image, and a photo of a settings toggle would not evidence anything. #269 has the full description and hardware details. I can record a video against a high refresh rate display if that would help.
image

## Breaking changes

None. Default behaviour is unchanged.

## Linked issues

Fixes #269. Related to #16.

### DIFF
--- a/composeApp/src/androidMain/kotlin/com/nuvio/app/Platform.android.kt
+++ b/composeApp/src/androidMain/kotlin/com/nuvio/app/Platform.android.kt
@@ -12,3 +12,7 @@ internal actual val isIos: Boolean = false
 internal actual val isDesktop: Boolean = false
 internal actual val isWindows: Boolean = false
 
+internal actual val isDesktopRenderBackendConfigurable: Boolean = false
+internal actual fun isDesktopOpenGlRendererEnabled(): Boolean = false
+internal actual fun setDesktopOpenGlRendererEnabled(enabled: Boolean) = Unit
+
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -1743,6 +1743,10 @@
     <string name="settings_playback_nvidia_rtx_video_section">NVIDIA RTX Video</string>
     <string name="settings_playback_nvidia_rtx_super_resolution">RTX Video Super Resolution</string>
     <string name="settings_playback_nvidia_rtx_super_resolution_desc">Upscale low-resolution video using NVIDIA RTX AI Super Resolution.</string>
+    <!-- Playback settings desktop display -->
+    <string name="settings_playback_desktop_display_section">Display</string>
+    <string name="settings_playback_desktop_opengl_renderer">Use OpenGL renderer</string>
+    <string name="settings_playback_desktop_opengl_renderer_desc">Draw the interface with OpenGL instead of Direct3D. Can reduce flicker or stutter on some high refresh rate displays. Restart the app to apply.</string>
     <!-- Playback settings iOS video output -->
     <string name="settings_playback_ios_video_output">iOS VIDEO OUTPUT</string>
     <string name="settings_playback_ios_hardware_decoder">Hardware decoder</string>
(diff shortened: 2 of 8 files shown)


he3als:
Can confirm that setting `SKIKO_RENDER_API=OPENGL` fixed it for my friend with an NVIDIA GPU.

As a workaround before this is merged, you can use a batch script to launch Nuvio:
```bat
set "SKIKO_RENDER_API=OPENGL" && start "" "%programfiles%\Nuvio\Nuvio.exe"
```

Issue #252, opened 2026-07-18
[Bug]: Whole app goes black when I play any source or any movie/show


Stefma4Git:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.1.13

### Install method

GitHub Release desktop installer/package

### Platform

Windows (desktop)

### Desktop hardware

Windows 11

### OS version

Windows 11

### Area (tag)

Playback (start/stop/buffering)

### Bug description

When I play something the whole screen goes black and I cant click on anything inside the app, I tried updating/reinstalling Nuvio

[Video showcasing the bug](https://drive.google.com/file/d/16Sa8a_Uw2dekoDa6Qqb0bObz_nFl-gEF/view?usp=sharing)

### Steps to reproduce

1. Open Nuvio
2. Select a show/Movie
3. select any source
4. get black screen

### Expected behavior

Movie/Show plays and I see UI player controls

### Actual behavior

When I select the source the whole screen goes black, I cant see or interact with player UI

### Frequency

Always

### Did this work before?

Yes, it used to work

### Logs (required for crash reports)

```shell

```

Stefma4Git:
there was also some UI bug in the VIDEO attached in the Home screen when I opened the app, idk what that is, but its unrelated, this black screen also happens unrelated to that Home screen bug

Poshfroshy:
This is also happenning here, works perfectly on mobile just my laptop, win 11

piyushxgoyal:
this i shappening for me also

PR #276, opened 2026-07-21
fix(player): stop the player leaving you on a blank screen with no way out

### LINKED ISSUE #270: [Bug]: Player route can strand the user on a blank screen when its launch payload is gone
### Pre-flight checks

- [ ] I can reproduce this on the latest desktop release or latest main build.

### App version / commit

0.1.13-alpha

### Install method

Built from source

### Platform

Windows

### Bug description

Note up front: this was found by code inspection while investigating #252, not from a user report, and I do not have reliable reproduction steps. Filing it because the failure mode is unrecoverable for the user if it is hit.

When `PlayerRoute` is composed with a `launchId` that is no longer in `PlayerLaunchStore`, it renders an empty full-screen `Box` and relies on the guarded pop to leave the screen. That pop can silently do nothing:

- `popHandled` is already spent if this route has popped before, so the guarded call is a no-op, and
- `popBackStack` also refuses when the player is the only entry on the back stack.

In either case nothing pops and nothing is drawn, so the user is left on a blank screen with no way out.

### Steps to reproduce

No reliable repro. Requires the route to be composed after its launch payload has been disposed.

### Expected behavior

If the launch payload is missing, the player route leaves the screen and returns the user somewhere usable.

### Actual behavior

Potentially renders a blank, unrecoverable screen.

### Additional context

The escape path could pop directly rather than through the spent guard, and fall back to the tabs root when the back stack cannot pop at all.

Laskco:
## Summary

Makes sure the player screen always sends you back somewhere usable, instead of occasionally leaving you stuck on a blank screen.

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

The player screen needs its playback data (the "launch payload") to show anything. If that data has already been cleaned up by the time the screen is shown, the screen draws nothing at all, and it relies on automatically sending you back to get you out.

The problem is that the auto-back can quietly do nothing, for two reasons:

1. It only fires once. If this screen has already gone back before, the guard that prevents double-navigation has been used up, so the call does nothing.
2. It refuses to go back when the player is the only screen on the stack.

If either happens, nothing navigates and nothing is drawn, so you are left staring at a blank screen with no way out (#270).

The fix makes the exit unconditional: go back directly rather than through the used-up guard, and if the stack genuinely cannot go back, go to the main tabs screen instead. Either way you end up somewhere usable.

## Desktop scope

This is in shared code (`App.kt`) so it applies to all platforms, but it matters for desktop: it was found while investigating the desktop black-screen reports in #252. To a desktop user, a blank unrecoverable screen looks exactly like the freeze described there.

## Issue or approval

Fixes #270

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Eleven lines, in one place: the branch that runs when the playback data is missing.

The shared back-navigation helper is not touched, so every other screen keeps its current behaviour. The blank placeholder is left alone, since it is only on screen for the moment before the route exits.

## Testing

Windows 11, built from source, 0.1.13-alpha.

Being upfront: I found this by reading the code while investigating #252, not from a real occurrence, and I do not have steps to reproduce it. So I cannot claim to have watched this fix rescue a stuck screen in practice.

What I did verify is that normal use is unaffected: entering and exiting playback, switching sources and episodes, and backing out all behave exactly as before. That is expected, because the changed branch is only reached when the playback data is already gone.

If you would rather wait for a reproduction before taking this, that is completely reasonable.

Not tested on macOS or Linux.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #270. Found while investigating #252.

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/App.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/App.kt
@@ -3017,8 +3017,18 @@ private fun MainAppContent(
                     )
                     val launch = remember(route.launchId) { PlayerLaunchStore.get(route.launchId) }
                     if (launch == null) {
+                        // The launch payload is already disposed, so this entry draws nothing but a
+                        // full-screen blank. The guarded pop cannot be trusted to get us out of it:
+                        // it no-ops once this route has popped before (popHandled is already spent),
+                        // and popBackStack also refuses when this is the only entry left. Either way
+                        // the user is stranded on an unrecoverable blank screen. Escape directly and
+                        // fall back to the tabs root if the stack itself cannot pop.
                         LaunchedEffect(route.launchId) {
-                            onBack()
+                            if (!navController.popBackStack(expectedRoute = route)) {
+                                navController.navigate(TabsRoute) {
+                                    popUpTo<PlayerRoute> { inclusive = true }
+                                }
+                            }
                         }
                         Box(modifier = Modifier.fillMaxSize())
                         return@entry


Laskco:
Rebased onto current Dev.

Flagging something that landed in the meantime: d514468f ("fix(navigation): defer route cleanup until pop completes") tackles the same underlying problem from the other end, by not clearing the payload until the pop has finished. That is the better fix, and it should make this situation much rarer.

It does not fully overlap though. The escape path in the `launch == null` branch is unchanged on Dev:

```kotlin
if (launch == null) {
    LaunchedEffect(route.launchId) { onBack() }
    Box(modifier = Modifier.fillMaxSize())
    return@entry
}
```

`onBack()` is still the guarded pop, so if that branch is ever reached with the guard already spent, or with the player as the only entry on the stack, nothing navigates and the user is still stuck on a blank screen. This PR only hardens that escape; it does not change when cleanup happens.

So the two are complementary rather than competing. That said, if you consider the failure mode now sufficiently unreachable in practice, closing this is a perfectly reasonable call and I will not argue it, especially as I still have no reproduction.

PR #277, opened 2026-07-21
feat(desktop): add Linux native player support

KhooLy:
## Summary
Adds native video playback on Linux. Same approach as macOS/Windows: mpv inside a GTK/X11 window, using the same NativePlayerBridge/NativePlayerController code. Added a Gradle task to build the native bridge on Linux.

## PR type
- [x] Approved larger or directional change

## Why
Linux has no player right now, just a "not supported" message. This adds the same native player Mac and Windows already have.

## Desktop scope
Linux only. Two shared files changed, both gated behind `DesktopHostOs.LINUX`:
- PlayerEngine.desktop.kt: dispose the controller when the host becomes non-displayable, otherwise the player gets stuck after closing it once.
- NativePlayerHost.kt: added a ComponentListener to detect first paint, since AWT's paint() doesn't reliably fire on Linux's GTK/X11 embedding.

## Issue or approval
Approved in #27

## UI / behavior impact
- [x] Behavior change has explicit maintainer approval

No UI change.

## Policy check
- [x] I have read and understood `CONTRIBUTING.md`.
- [x] This PR is small, focused, and limited to one problem.
- [x] This PR is scoped to the desktop app, desktop packaging, desktop documentation, or shared code required for desktop behavior.
- [x] This PR is not cosmetic-only.
- [x] Any UI change fixes a linked glitch/bug and includes visual proof, or this PR has no UI change.
- [x] Any behavior change fixes a linked bug/regression or has explicit approval, or this PR has no behavior change.
- [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [x] This PR does not add dependencies, architecture changes, migrations, or product-direction changes without explicit approval.
- [x] I listed the testing performed below.

## Scope boundaries
Playback only, no packaging changes.

## Testing
Tested manually on CachyOS, X11/XWayland, Intel iGPU + Nvidia RTX 4060 Laptop hybrid setup.
- Player controls work
- Confirmed hardware decode is active (mpv's hwdec-current)
- Closed and reopened the player repeatedly, no black screen or stuck state
- Mac/Windows paths unaffected, Linux code is gated behind DesktopHostOs.LINUX

## Screenshots / Video
Not a UI change

## Breaking changes
None

## Linked issues
Approved in #27

PR #295, opened 2026-07-27
Add Linux desktop support (native libmpv player bridge)

### LINKED ISSUE #27: [Feature]: Linux Support
### Area (tag)

### Problem statement

It would be awesome to have Linux Support with flatpak or AppImage releases

### Proposed solution

Have Linux support

No, this is only a proposal

_No response_

_No response_

ArjunKdaf:
## Summary

Native Linux support for the desktop app. A libmpv/GTK JNI bridge (`player_bridge.cpp`) renders video embedded in the app window, with the shared controls page composited over it via WebKitGTK snapshots — the same controls, behavior and flow as the macOS and Windows players. Includes Linux packaging (DEB with correct runtime dependencies), TorrServer bundling, and a player teardown barrier that releases the native player before navigation.

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

Linux users currently have no way to run Nuvio Desktop. This adds it, requested in #27 and approved for merge as experimental support by @skoruppa in this PR's thread.

## Desktop scope

Linux, plus shared desktop code where the Linux player needed a seam:

- `commonMain`/shared desktop: the player release barrier (by @ofers) runs on all desktop platforms; on macOS/Windows the dispose paths are bounded and EDT-independent, and defaults keep behavior unchanged there. Kotlin platform gates (`WINDOWS || LINUX`) opened for webview preload/warmup.
- Everything else is Linux-gated: the JNI bridge, GTK/WebKitGTK compositing, DEB packaging post-processing, TorrServer Linux binary.

## Issue or approval

Closes #27. Approved as experimental support by @skoruppa in this thread; scope checklist by @kali113 (rebase onto current Dev, teardown PR merged, template filled) is complete.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

No UI or behavior change on macOS or Windows beyond the no-op release barrier described above. On Linux, everything is new.

## Policy check

## Scope boundaries

Kept out on purpose, per the scope discussion in this thread: Flatpak/AppImage packaging (#324 tracks Flatpak), native Wayland (the player runs on XWayland), RPM packaging, and distro-specific desktop integration (desktop entry, AUR PKGBUILD — @posilash plans to raise those separately). No UI polish, no unrelated refactors.

## Testing

- Author, continuous: NixOS, KDE and GNOME, NVIDIA hybrid. Full playback flows, fullscreen, keyboard, controls overlay, teardown, TorrServer streams.
- JVM test suite on the merged branch: 123 suites / 744 tests / 0 failures. Bridge `.so` and the vendored compose-media-player native build both compile clean.
- Community testers in this thread: @Minwalin (Fedora KDE, RPM-built), @skoruppa (Intel i915), @FeelThePoveR (NVIDIA 4070 Super), @Paolo86cripple (AMD, Wayland session), @kali113 (Arch, GNOME Wayland, nouveau — `x11egl` path), @posilash (Arch, Hyprland, NVIDIA proprietary — `x11vk` fallback path plus `nvdec-copy` hardware decoding, 4K HEVC HDR), @shscs911 (daily driver for over a week), @ofers (DEB install, TV audio passthrough).
- DEB packaging verified by fixture tests (`verify-linux-deb.sh`).

## Screenshots / Video

Linux player running embedded: see the screenshots and tester reports earlier in this thread (KDE, GNOME, Hyprland).

## Breaking changes

None.


Minwalin:
This seems to overlap substantially with #277, since both add a native libmpv player bridge for Linux. Is this intended to replace that PR, or could the additional TorrServer, WebKitGTK overlay, packaging, and EOF changes be contributed on top of #277?

Minwalin:
### Environment

- OS: Fedora 44 KDE Plasma
- Session: Wayland
- Branch: `linux-desktop-support`

### Issue

The application compiles successfully and launches normally.

However, when starting it under a Wayland session, attempting to play a video causes the application to crash.

The log ends with:

```text
X Error of failed request: BadMatch (invalid parameter attributes)
Major opcode: 142 (Composite)
```

Gradle then reports:

```text
Task :composeApp:run FAILED
```

### Workaround

Launching the application with:

```bash
GDK_BACKEND=x11 ./gradlew :composeApp:run
```

makes the issue disappear.

Video playback works correctly and the application no longer crashes.

This seems related to the Linux native player/XComposite implementation running through Wayland/XWayland.
The native player crashes under KDE Plasma Wayland with an XComposite BadMatch error, but works correctly when forcing GDK_BACKEND=x11. Have you tested the Linux player under Plasma Wayland, or only under X11/GNOME Wayland?

FeelThePoveR:
My Nvidia RTX 4070 Super still has the same issues as with all the other PRs - GBM buffer creation failures and EGL make current failures causing black screen issues. 

Both of those were resolved on the Skoruppa's PR (that wasn't using webkit) by using GLX if Nvidia was detected (not sure if this can be applied for webkitgtk) or worked around by using Zink driver.

```
Failed to create GBM buffer of size 1220x780: Invalid argument
Failed to create GBM buffer of size 1220x780: Invalid argument
```

```
[nuvio-bridge] mpv[v] vo/gpu/opengl: EGL_VERSION=1.5

[nuvio-bridge] mpv[v] vo/gpu/opengl: EGL_VENDOR=NVIDIA

[nuvio-bridge] mpv[v] vo/gpu/opengl: EGL_CLIENT_APIS=OpenGL_ES OpenGL

[nuvio-bridge] mpv[v] vo/gpu/opengl: Trying to create Desktop OpenGL context.

[nuvio-bridge] mpv[v] vo/gpu/opengl: Choosing visual EGL config 0x18, visual ID 0x27

[nuvio-bridge] mpv[v] vo/gpu/x11: The XPresent extension was found.

[nuvio-bridge] mpv[v] vo/gpu/x11: XPresent disabled.

[nuvio-bridge] mpv[fatal] vo/gpu/opengl: Could not make context current!
```

ArjunKdaf:
@Minwalin

> This seems to overlap substantially with #277, since both add a native libmpv player bridge for Linux. Is this intended to replace that PR, or could the additional TorrServer, WebKitGTK overlay, packaging, and EOF changes be contributed on top of #277?

Thanks for flagging #277 — I developed this independently and hadn't seen it when I started, and credit to @KhooLy for getting there first. There's clear overlap in intent (both add a native libmpv bridge for Linux), so it's worth laying out how #295 is scoped for you and the maintainers to compare:

- **Interactive controls overlay** — the bulk of the work here. It renders Nuvio's *shared* `player-ui` controls page in a WebKitGTK layer composited over the video (X Composite extension + mpv `overlay-add`), so the player interface is identical to macOS/Windows: play/pause, scrubber, Subs/Audio/Sources, loading screen, fullscreen, keyboard shortcuts. From #277's diff (a `.c` bridge with no `player-ui`/WebKitGTK changes) it reads as video-only, so this is the main difference.
- **Bundled TorrServer** (`linux-amd64`, matching the committed macOS binaries) so P2P streaming works out of the box.
- **Decoder + behavior parity with the macOS bridge** — `hwdec=auto`/`gpu-hwdec-interop`, EOF via `eof-reached` (so next-episode and play/pause behave correctly), plus teardown crash-hardening and a Plasma-Wayland (`GDK_BACKEND=x11`) fix from tester feedback here.
- Packaging via the existing `TargetFormat` (deb/rpm).

Because the controls layer is a different architecture (a C++ WebKitGTK compositor vs #277's C bridge), folding these onto #277 would be closer to a rewrite than a patch, so I've kept #295 self-contained.

Happy to defer to the maintainers on which direction they prefer, and glad to coordinate with @KhooLy either way — whatever gets Linux users a working player soonest works for me.

ArjunKdaf:
@Minwalin — thanks for the detailed Plasma Wayland report, that was exactly the missing piece.

> X Error of failed request: BadMatch (invalid parameter attributes) — Major opcode: 142 (Composite) … Launching with `GDK_BACKEND=x11` makes the issue disappear.

Fixed in 3ac3ed6: the bridge now forces the X11 GDK backend (`gdk_set_allowed_backends("x11")` before `gtk_init`), so the controls window is always an X11 surface that XComposite can redirect — i.e. your `GDK_BACKEND=x11` workaround is baked in. Would you mind re-testing on Plasma Wayland **without** the env var? It should just work now.

ArjunKdaf:
@FeelThePoveR — thanks, the log pins it: not decode, it's Nvidia's EGL/GBM context creation failing under `x11egl` (`Could not make context current`).

GLX-if-Nvidia won't work here, and your "may not apply to webkitgtk" hunch is why: the controls are a WebKitGTK layer composited over the video via mpv `overlay-add`, which needs the EGL path. Forcing `gpu-context=x11` (GLX) kills that compositing (`display depth: -1`) → video works but controls vanish. Worse trade.

Zink keeps EGL intact, and you mentioned it as one of the workarounds on the other PR — could you test it here with no code change?

```
MESA_LOADER_DRIVER_OVERRIDE=zink GALLIUM_DRIVER=zink <run cmd>
```

If you get video **and** working controls, I'll add Nvidia→Zink detection (or document it) — no regression for anyone. What's your driver version?

Minwalin:
I've done some more testing on Fedora 44.

The X11 changes in this PR worked well on my system. Video playback is now working correctly without needing to force `GDK_BACKEND=x11`, so this definitely fixes the playback issue I was experiencing.

I also investigated a separate audio issue I was having.

MPV correctly detected the audio stream:

```text
Audio --aid=1 --alang=kor (eac3 6ch 48000 Hz) [default]
```

but then failed with:

```text
ad: Failed to initialize a decoder for codec 'eac3'
Audio: no audio
```

It turned out this wasn't caused by Nuvio itself. My Fedora installation only had the default FFmpeg packages, which don't include the E-AC-3 decoder. After installing the full multimedia codec support (RPM Fusion), `ffmpeg -decoders | grep eac3` showed:

```text
A....D eac3  ATSC A/52B (AC-3, E-AC-3)
```

After restarting Nuvio, audio playback worked correctly.

So from my testing, the X11 changes in this PR are working as expected, and the audio issue was caused by missing system codecs rather than the Linux player implementation.

The only remaining issue I've noticed is that the audio track selector still shows **"No audio tracks available"** even when MPV detects the track and playback works, so that appears to be a separate issue.

Minwalin:
Just as an additional note, I also noticed a crash that was already present before this PR, so I don't think it's related to these changes.

If I leave the application idle for a while, it occasionally crashes on exit with a SIGABRT.

The top of the stack trace is:

```text
Signal: 6 (SIGABRT)

WebKit::WebsiteDataStore::~WebsiteDataStore()
webkit_website_data_manager_finalize()
webkit_web_context_finalize()
```

From the backtrace, it looks like the abort happens during WebKitGTK cleanup/shutdown rather than in the MPV playback code.

I just wanted to mention it in case it's helpful for future debugging. I can provide the full coredump if needed.

ArjunKdaf:
@Minwalin — thanks, both fixed on the branch:

- **Audio/subtitle selectors** (e19b4bb): the bridge fed the controls empty track arrays + raw mpv `track-list` the Kotlin decoder can't parse. Now builds the formatted per-type list like macOS — verified Audio shows the DTS-HD track, Subs the English subtitle, both selectable.
- **SIGABRT on exit** (c5f12d5): the controls webview used the default *persistent* web context, whose `WebsiteDataStore` finalizes at exit while the GTK thread still runs. Switched to an ephemeral context (the page needs no persistence). Intermittent, so a re-test would confirm.

FeelThePoveR:
@ArjunKdaf 
The issue here isn't with the mpv - I got mpv working on GLX with GTK UI "visible" with GTK cairo software rendering for example, but I couldn't get the transparency to work in any way when either disabling dmabuf or running with cairo so the playback was obscured when UI was visible.
Those 2 logs are coming from webkitgtk not from mpv
```
Failed to create GBM buffer of size 1220x780: Invalid argument
Failed to create GBM buffer of size 1220x780: Invalid argument
```

But anyway running the app with `MESA_LOADER_DRIVER_OVERRIDE=zink GALLIUM_DRIVER=zink` on this PR results in
```
Debug: (NativePlayerControls) control callbacks attached handle=0
Debug: (NativePlayerControls) play handle=0
Debug: (NativePlayerControls) attach requested source=scheme=https length=847 hash=-314082171 headers=0 playWhenReady=true initialPositionMs=828036 decoderPriority=1
[nuvio-bridge] mpv_initialize failed: invalid parameter
Warn: (NativePlayerControls) attach failed source=scheme=https length=847 hash=-314082171
java.lang.IllegalStateException: Native player did not return a handle.
        at com.nuvio.app.features.player.desktop.NativePlayerController.createPlayer$lambda$4(NativePlayerController.kt:159)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

Running it with `__GLX_VENDOR_LIBRARY_NAME=mesa MESA_LOADER_DRIVER_OVERRIDE=zink GALLIUM_DRIVER=zink` fails a little bit further with GBM buffer creation errors
```
Failed to create GBM buffer of size 890x714: Invalid argument
Failed to create GBM buffer of size 890x714: Invalid argument
```

So the PR in this form doesn't work with Zink either. I got Zink to work on GTK4/webkitgtk6.0 when I tinkered around with it a month ago here: https://github.com/FeelThePoveR/NuvioDesktop/pull/1 (no success on normal Nvidia driver hence I dropped it in hopes that somebody else will manage to get it working). The code there is half-vibecoded and messy since I didn't see a point in getting this merged upstream as it didn't work as well as I would have liked, but maybe it'll give you some insight.

Minwalin:
> @Minwalin — thanks, both fixed on the branch:
> 
>     * **Audio/subtitle selectors** ([e19b4bb](https://github.com/NuvioMedia/NuvioDesktop/commit/e19b4bb54ba86f4a56fb1fac24c20098e4dec3e7)): the bridge fed the controls empty track arrays + raw mpv `track-list` the Kotlin decoder can't parse. Now builds the formatted per-type list like macOS — verified Audio shows the DTS-HD track, Subs the English subtitle, both selectable.
> 
>     * **SIGABRT on exit** ([c5f12d5](https://github.com/NuvioMedia/NuvioDesktop/commit/c5f12d5b0c7008991b4c42ff68429d9af15ef780)): the controls webview used the default _persistent_ web context, whose `WebsiteDataStore` finalizes at exit while the GTK thread still runs. Switched to an ephemeral context (the page needs no persistence). Intermittent, so a re-test would confirm.

The latest changes definitely improved the native player, but I found two remaining issues with audio track handling on Linux.

### 1. Audio track metadata is displayed incorrectly

The audio tracks are now detected and listed, but some metadata is incorrect.

Example:

```
(DKS) LINE (stereo, ac3)
(FS) CAM (unknown2, ac3)
(LMHD) LINE (unknown2, ac3)
(????) RETAIL (unknown2, ac3)
```

Some tracks show `unknown2` instead of the channel layout, and some language names appear as `????`.

### 2. Selecting another audio track has no effect

The audio tracks are listed, but selecting a different one does not actually change the audio.

The terminal logs show that the event reaches the native player, but it is never handled:

```
Debug: (NativePlayerControls) event received handle=140225851397152 type=selectAudioTrack value=2.0
Debug: (PlayerControls) event type=selectAudioTrack value=2.0 video=tt30825738 s=- e=- pos=132458 duration=7918084 speed=1.0 controller=true
Debug: (NativePlayerControls) event delegated type=selectAudioTrack handled=false
```

The same happens with other tracks:

```
Debug: (NativePlayerControls) event received handle=140225851397152 type=selectAudioTrack value=3.0
Debug: (PlayerControls) event type=selectAudioTrack value=3.0 video=tt30825738 s=- e=- pos=137042 duration=7918084 speed=1.0 controller=true
Debug: (NativePlayerControls) event delegated type=selectAudioTrack handled=false
```

It looks like the UI correctly sends the `selectAudioTrack` event, but the native player/controller returns `handled=false`, so the request never reaches mpv.

Environment:
- Fedora 44 KDE Plasma (Wayland)
- Latest `linux-desktop-support` branch
- Native Linux player

ArjunKdaf:
@FeelThePoveR — I reproduced your black screen by PRIME-offloading to the Nvidia GPU on my hybrid laptop, and it's **two stacked issues**, both now addressable:

1. **mpv can't make an `x11egl` context current** on the embedded (foreign AWT `wid`) window — total black. `x11vk` (Vulkan) initializes fine there. I pushed an env override (1a1e0e3): `NUVIO_MPV_GPU_CONTEXT=x11vk`.
2. **The controls overlay** fails because WebKit's DMABUF renderer window has no XComposite-redirectable pixmap (`BadPixmap` on `NameWindowPixmap`) — those are your GBM errors. `WEBKIT_DISABLE_DMABUF_RENDERER=1` fixes it (0 errors, controls composite).

With both set, I get **working controls + Vulkan video decode** on Nvidia. The only thing I *can't* validate is the final picture: on my hybrid, the PRIME render-offload copy to the Intel display corrupts the video (centered-square/tearing). Your pure-Nvidia desktop presents directly, so that's almost certainly my offload artifact, not yours.

Could you pull 1a1e0e3 and test:
```
NUVIO_MPV_GPU_CONTEXT=x11vk WEBKIT_DISABLE_DMABUF_RENDERER=1 ./gradlew :composeApp:run
```
and report whether video is clean + controls work? If it's good on real Nvidia, I'll wire auto-detection so it's automatic.

ArjunKdaf:
@Minwalin — both fixed (and I confirmed audio switching works locally):

1. **Track switching** (649a2ad): the controls webview's `selectAudioTrack` event had no case in the shared player-controls dispatch (only subtitles did), so it returned `handled=false` and never reached mpv. Added the handler.
2. **Channel labels** (41a0a9a): `unknown2` is mpv's name for an *unknown 2-channel layout* — now mapped from the channel count to Mono/Stereo/5.1/7.1 like the macOS bridge.

The `????` is the file's own track title (a scene-group tag baked into the mkv), not something the player generates — so that one's just the source metadata.

Both need a rebuild to pick up — mind re-testing?

FeelThePoveR:
@ArjunKdaf 
Running the app with NUVIO_MPV_GPU_CONTEXT=x11vk WEBKIT_DISABLE_DMABUF_RENDERER=1 ./gradlew :composeApp:run aligns with how it worked in my previous testing.
It "works" as in the UI shows up, but the transparency on it fails so it obscures the playback whenever it's visible. It also leaves significant ghosting artifacts from previous ui elements on the non-transparent background (2 loaders on the screenshot below). 

In general from what I've seen disabling the webkitgtk dmabuf or running it with cairo software renderer is a no-go for that reason.

The good part here is that mpv initializes just fine with x11vk.

image

ArjunKdaf:
@FeelThePoveR — the opaque overlay + ghosting you hit is exactly what made me drop the pixmap-readback approach entirely. The root cause is that compositing by reading the redirected controls window's X pixmap is renderer-dependent, and on Nvidia *neither* renderer works: the default DMABUF renderer leaves the pixmap empty (your `Failed to create GBM buffer` errors → controls vanish), while disabling it (`WEBKIT_DISABLE_DMABUF_RENDERER=1`, or cairo) makes WebKit paint the page background **opaque** into the pixmap — which is the obscured playback and the stale-loader ghosting in your screenshot. Same conclusion you reached in your GTK4 experiment: that path is a dead end on Nvidia.

So it's reworked in 79492be: the bridge now captures the controls with async `webkit_web_view_get_snapshot(…, TRANSPARENT_BACKGROUND)`, which hands back premultiplied ARGB with real alpha straight from the page, independent of renderer and driver. DMABUF stays enabled, no env var needed.

And since you confirmed x11vk initializes fine on your card, e43e6ee wires the auto-detection: the bridge attempts `x11egl` first, and when `mpv_initialize` fails (Nvidia's EGL refusing to make a context current on the foreign AWT window), it retries with `x11vk` paired with `nvdec-copy` — Vulkan's *native* hwdec interop is what produced the corrupt frames, copy-back NVDEC is clean. The `NUVIO_MPV_GPU_CONTEXT`/`NUVIO_MPV_HWDEC` overrides still exist to pin a single attempt for testing, but nothing needs them anymore.

Verified locally on both paths of my hybrid laptop, zero env vars:

- **Intel iGPU** — `x11egl` succeeds first try, controls transparent, video clean.
- **PRIME offload to the RTX 3070** (Nvidia EGL fails exactly like your desktop) — auto-falls back to `x11vk` + `nvdec-copy`: clean video (the corruption I reported earlier was the Vulkan interop, now avoided), transparent controls over playback, no ghosting, no GBM errors.

Could you pull the latest branch and run plain `./gradlew :composeApp:run` — no env vars — on the 4070 Super? If pure-Nvidia is clean too, I think that closes out the Nvidia thread.

FeelThePoveR:
@ArjunKdaf 
The current approach behaves weirdly GBM buffer creation still fails but the UI renders although it's opaque/semi-transparent and either persist or disappear which is also inconsistent, but most of the time it lingers and doesn't hide. The UI state lags behind on load as well - the loader isn't spinning unless you interact with the UI.

Logs
```
Debug: (NativePlayerControls) attach requested source=scheme=https length=892 hash=1923033149 headers=0 playWhenReady=true initialPositionMs=902568 decoderPriority=1
[nuvio-bridge] mpv_initialize failed (gpu-context=x11egl): invalid parameter
[nuvio-bridge] gpu-context x11egl failed to initialize; using x11vk instead
Debug: (NativePlayerControls) attach created handle=140337184819728 source=scheme=https length=892 hash=1923033149 initialPositionMs=902568
Debug: (NativePlayerControls) applied remembered volume level=1.0 handle=140337184819728
Debug: (NativePlayerControls) updateControls handle=140337184819728 title=The Boys pos=0 duration=0 speed=1x audioLabel=Audio subsLabel=Subs fullscreen=false
Failed to create GBM buffer of size 1920x1006: Invalid argument
Failed to create GBM buffer of size 1920x1006: Invalid argument
```
Screenshot
image

Out of curiosity I tried it with Zink as well and it behaves the same as on the Nvidia driver,

FeelThePoveR:
@ArjunKdaf 
Alright another test run this time with `WEBKIT_DISABLE_DMABUF_RENDERER=1` and this approach works with it just fine. The UI renders correctly and the background is fully transparent like it should be so setting this in the bridge for Nvidia should be good. There's an issue with flickering during window drag, but it's a pretty small issue all things considered.

ArjunKdaf:
@FeelThePoveR — great, that confirms the fix. I'll make the bridge set `WEBKIT_DISABLE_DMABUF_RENDERER=1` automatically on Nvidia so no env vars are needed, and I'll investigate the window-drag flickering. Really appreciate all your testing — it's what got Nvidia working.

skoruppa:
I am impressed :) I see you managed to encounter all the same issues I did in my attempts to bridge Linux, and it seems you did a way better job than me dealing with them ;)

One thing that I always have to deal with (and it seems to be somehow connected to Ubuntu 26.04 with Gnome and Intel GPU) is:

```
Debug: (NativePlayerControls) control callbacks attached handle=0
Debug: (NativePlayerControls) play handle=0
Debug: (NativePlayerControls) attach requested source=scheme=https length=298 hash=-270298161 headers=0 playWhenReady=true initialPositionMs=141708 decoderPriority=1

(java:2031258): GLib-GObject-CRITICAL **: 12:15:03.956: cannot register existing type 'GdkDisplayManager'

(java:2031258): GLib-CRITICAL **: 12:15:03.956: g_once_init_leave: assertion 'result != 0' failed

(java:2031258): GLib-GObject-CRITICAL **: 12:15:03.956: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
```
when trying to launch the player. In my changes I always had to add `initGtkEarly` to `main()`  (essentially calling `gtk_init_check` via `dlopen` BEFORE any AWT/Compose/Skia initialization)

```
fun main(args: Array<String>) {
    // On Linux, initialize GTK BEFORE AWT/Compose/Skia to prevent GdkDisplayManager
    // type registration conflict (Skiko partially loads GDK without full GTK init).
    if (System.getProperty("os.name", "").lowercase().contains("linux")) {
        runCatching { NativePlayerBridge.initGtkEarly() }
    }
    // ... rest of init
}
```
with native side:
```
JNIEXPORT jboolean JNICALL
Java_com_nuvio_app_features_player_desktop_NativePlayerBridge_initGtkEarly(
    JNIEnv *env, jclass cls) {
    (void)env; (void)cls;
    void *gtk_lib = dlopen("libgtk-3.so.0", RTLD_NOW | RTLD_GLOBAL);
    if (!gtk_lib) return JNI_FALSE;
    typedef int (*fn_gtk_init_check_t)(int*, char***);
    fn_gtk_init_check_t init_fn = (fn_gtk_init_check_t)dlsym(gtk_lib, "gtk_init_check");
    if (!init_fn) return JNI_FALSE;
    int argc = 0;
    if (!init_fn(&argc, NULL)) return JNI_FALSE;
    return JNI_TRUE;
}
```
If you have better idea on how to handle this, be my guest 

Other issues I noticed
- I sometimes have no cursor. When it is present I can click on things, but there is no "hover" action and sometimes I need to press things twice. I can do a "scroll" action on my mouse to update cursor position and show my cursor or to hover over things :P 
- You can right-click to invoke webkit menu 
- Overlay does not follow window size. If I change window size or switch to full screen, player overlay will stay the same size
- Sometime overlay works fine and sometime I see this semi opaque box

obraz
and if that state occur I need to use `WEBKIT_DISABLE_DMABUF_RENDERER=1` for it to work (and it seems to work better for me with DMABUF disabled)

FeelThePoveR:
> @FeelThePoveR — great, that confirms the fix. I'll make the bridge set `WEBKIT_DISABLE_DMABUF_RENDERER=1` automatically on Nvidia so no env vars are needed, and I'll investigate the window-drag flickering. Really appreciate all your testing — it's what got Nvidia working.

No problem. I've been invested into getting this working for a while now and when I looked at your approach I went "huh, this approach looks different enough from what I or the others have been doing that it might just work out" so really big thanks and congratulations to you for solving this conundrum known as "Nvidia on Linux"  :D

Minwalin:
> @Minwalin — both fixed (and I confirmed audio switching works locally):
> 
>     1. **Track switching** ([649a2ad](https://github.com/NuvioMedia/NuvioDesktop/commit/649a2ad7c307c63cb87a79372efc6e4b0f9d0231)): the controls webview's `selectAudioTrack` event had no case in the shared player-controls dispatch (only subtitles did), so it returned `handled=false` and never reached mpv. Added the handler.
> 
>     2. **Channel labels** ([41a0a9a](https://github.com/NuvioMedia/NuvioDesktop/commit/41a0a9ae53085bfe66d8df94fb887a3faf2fe28b)): `unknown2` is mpv's name for an _unknown 2-channel layout_ — now mapped from the channel count to Mono/Stereo/5.1/7.1 like the macOS bridge.
> 
> 
> The `????` is the file's own track title (a scene-group tag baked into the mkv), not something the player generates — so that one's just the source metadata.
> 
> Both need a rebuild to pick up — mind re-testing?

I rebuilt the latest release RPM using the portable JDK, reinstalled it, and tested it again. Audio track switching is now working correctly on Fedora Linux. Thanks for the fix!

From my testing so far, I think the Linux version is in a good state for its first public release. It would be great to let more Linux users test it and provide feedback.

ArjunKdaf:
Thanks a lot for the kind words @skoruppa, and for the detailed test report — it was really useful. To chase your findings I set up an Ubuntu 26.04 testbed VM (GNOME, stock OpenJDK 17, tested both with my RTX 3070 passed through via VFIO and on plain Mesa/virtio), and also validated everything on GNOME on my NixOS machine directly. Everything on your list traced back to four root causes. All fixed and pushed to this branch:

- **Frozen fades / stuck dark scrim / missing loading & pause screens** — the composite-redirected overlay window is never presented, so its GTK frame clock stalls under mutter (and KWin): the page's CSS animations froze mid-flight. Fixed by requesting a frame-clock update phase on every composite tick.
- **Cursor / hover / double-click issues** — two bugs stacked: the unconditional XRaiseWindow every 200ms made mutter's XWayland emit pointer crossing events each tick (endless fake cursor activity, chrome never auto-hid, hover thrashed) — it now only raises when not already topmost; and the overlay window defined no cursor of its own, so the pointer went blank over the controls — it now sets one explicitly.
- **Semi-opaque box** — you were right that `WEBKIT_DISABLE_DMABUF_RENDERER=1` helps: the dmabuf renderer degrades snapshot alpha on Mesa too (an alpha histogram of the snapshots showed zero fully-transparent pixels), not just on NVIDIA. It's now disabled unconditionally for the controls webview (overwrite=0, so an explicit env setting still wins).
- **Right-click menu** — suppressed, matching the macOS/Windows player webviews.

On initGtkEarly and the GdkDisplayManager crash: I couldn't reproduce it with this branch on the stock Ubuntu 26.04 JDK in the VM (clean start, zero GLib criticals). But your fix is sound defensive hardening and costs nothing, so I adopted it — gtk_init on the JVM main thread before AWT/Skia plus `-Djdk.gtk.version=0`, with one change: it pins `gdk_set_allowed_backends("x11")` first, since the overlay is driven through X11/XComposite and must not come up on the Wayland backend. I credited you with a Co-authored-by on that commit — I hope that's OK; if you'd rather not be on it, tell me and I'll remove it. Your snapshot watchdog idea is in as well (a `get_snapshot` requested before the page starts loading never calls back and froze the overlay for the whole session) — also credited in the commit message.

Would you mind retesting the branch on your machine? Curious whether cursor/hover, the fades, and the semi-opaque box are gone for you too — and whether resize/fullscreen tracking now behaves.

skoruppa:
The resize issue is still there
obraz
I noticed that after window resizing (or going full screen) the UI no longer updates (it will not change the time for example, will not hide anymore). 

The mouse cursor issues is also still there - mouse can disappear and items are not marked when I hoover over them. If I use my mouse scroll an item I have my mouse over will light up, and my cursor will be displayed. 

This looks like my cursor position is not actively tracked? Missing mouse is probably mpv behavior that hides mouse when there is no movement for some time (and as cursor position is not correctly tracked it just see it as not moving). And doing a scroll action "updates" my cursor position (just like clicking)

Minwalin:
I found another Linux-specific issue while testing the latest desktop Linux branch.

When starting video playback, the loading screen is missing the backdrop artwork and media title.

### Expected behavior

The loading screen should display:
- Backdrop artwork
- Media title
- Loading spinner

### Actual behavior

On Linux, only the loading spinner is displayed on a black background. The backdrop artwork and title are not shown.

Playback itself works correctly once loading finishes, so the issue appears to be limited to the loading screen before playback starts.

### Comparison

**Windows (expected):**
- ✅ Backdrop artwork
- ✅ Media title
- ✅ Loading spinner

**Linux (actual):**
- ❌ Backdrop artwork
- ❌ Media title
- ✅ Loading spinner

### Environment

- Fedora 44 KDE Plasma
- Wayland
- Linux desktop branch
- RPM build

I've attached screenshots comparing the Windows and Linux loading screens.

This on linux (Fedora)
cargaLinux

This is in windows 10
cargawindows

ArjunKdaf:
Update: seven more fixes pushed (`0af1cb7`..`879fa06`), verified on both GNOME/mutter and KDE/KWin.

**@skoruppa — hover/cursor root-caused, and this should fix your i915 box too.** It's not driver-specific: GDK compresses motion events by deferring them to the frame clock's flush phase, and mutter's X frame-sync never completes a frame for the composite-redirected, never-presented overlay — so motion events were silently dropped (~1 per pointer crossing). Buttons/scroll aren't compressed, which is exactly why click/scroll still worked and woke the cursor on your machine, and KWin has no frame-sync, which is why KDE never showed it — your symptom profile matches this bug exactly. Fix: `gdk_window_set_event_compression(FALSE)` on the overlay (`d6fc768`) — motion delivery went from ~0% to ~100%. Related fixes: the cursor now hides/shows natively with the chrome (`e5e6713` — the page's CSS `cursor:none` can't propagate through the redirected overlay), fullscreen exit no longer gets stuck under mutter (`13c5054` — Compose skips the exit write when the WM's extra state events make it think it's already windowed; now enforced against the AWT device), and Esc leaves the player like on macOS/Windows (`879fa06`).

**@skoruppa — resize/fullscreen freeze fixed** (`0af1cb7`, `d748571`): the resize relied on `gtk_window_resize`, which runs on that same stalled frame clock — it never landed and the composite loop early-returned forever. Now the X window is resized directly with a synchronous allocation, plus stale-size snapshots are dropped mid-resize — should also cover @FeelThePoveR's drag flicker.

**@Minwalin — loading screen fixed** (`9fa821b`): the metadata push arrived before the webview existed and was dropped (Kotlin dedups and never resends). The bridge now buffers and retries until the page confirms delivery, like the macOS/Windows bridges — full artwork + logo + stats.

**Known issues I'm leaving out of this PR** (keeping it scoped to Linux playback parity while the #295/#277 decision is pending):
- ~1.5–2s black gap between the Compose loading screen and the webview one — WebKitGTK cold start; Windows hides it with a webview pre-warm the bridge doesn't implement yet. Needs a Linux warmup path; candidate follow-up.
- While playing fullscreen, the overlay keeps compositing transparent snapshots (the hide signal that disengages it windowed never fires in fullscreen). Pure efficiency, no visible effect; the fix touches the regression-tested composite path, so deferring.

Minwalin:
> **@Minwalin — loading screen fixed** (`9fa821b`): the metadata push arrived before the webview existed and was dropped (Kotlin dedups and never resends). The bridge now buffers and retries until the page confirms delivery, like the macOS/Windows bridges — full artwork + logo + stats.
> 
> **Known issues we're leaving out of this PR** (keeping it scoped to Linux playback parity while the #295/#277 decision is pending):
> - ~1.5–2s black gap between the Compose loading screen and the webview one — WebKitGTK cold start; Windows hides it with a webview pre-warm the bridge doesn't implement yet. Needs a Linux warmup path; candidate follow-up.
> - While playing fullscreen, the overlay keeps compositing transparent snapshots (the hide signal that disengages it windowed never fires in fullscreen). Pure efficiency, no visible effect; the fix touches the regression-tested composite path, so deferring.
> 

Confirmed on Fedora 44 KDE.

The loading screen now correctly shows the artwork, logo, title, and metadata. The issue I reported is fixed.

I still notice a brief (~1 second) Compose loading screen with the spinner before the WebView loading screen appears, but that matches the known issue you mentioned about the WebKitGTK cold start.

Thanks for the quick fix!

skoruppa:
@ArjunKdaf this is the best state the bridge ever have been ;) 

I managed to get UI stuck few times, and when that was going video playback was really choppy, in logs I saw
```
[nuvio-bridge] snapshot stuck in flight >30 ticks, resetting (watchdog)
[nuvio-bridge] snapshot stuck in flight >30 ticks, resetting (watchdog)
[nuvio-bridge] snapshot stuck in flight >30 ticks, resetting (watchdog)
[nuvio-bridge] snapshot stuck in flight >30 ticks, resetting (watchdog)
```
and after that everything went back to normal. 

About the missing warmup, I think that without it if player starts fast enough it does not receive user's data and content's meta (the player is on default theme, not the one I picked and there are no titles/episodes list)

Sometimes mouse movement does not revel the hidden ui

I'm also not sure how fast the UI should render/refresh? Right now if I quickly hover over various buttons I see they light up with really slight but visible delay, but maybe I'm just finding nitpicking ;)


It works way better than #277 bridge

ArjunKdaf:
@skoruppa thanks for the thorough retest — glad the bridge is in its best state yet. All four of your reports are addressed in the 5 commits I just pushed (6f57753..b954df6):

**Fast-start metadata gap (default theme, no titles/episode list)** — this one turned out not to be a warmup/timing issue: the Linux bridge converted JNI strings with `GetStringUTFChars`, which returns *Modified* UTF-8 (CESU-8). Any non-BMP character — like the 👤/💾 emoji Torrentio puts in stream titles — produced bytes WebKit rejects, so the whole controls eval silently failed, and since the Kotlin side dedups payloads by structure it never resent. A fast start correlates because a warm session collapses to exactly one push: one loss = permanent. Fixed by converting from UTF-16 like the mac/win bridges (6f57753), and the payload is now also kept durable and re-delivered if the web process ever crashes and reloads (1a62cab).

**UI stuck + choppy video + the "snapshot stuck in flight" storms** — the watchdog itself had a real bug: it abandoned stalled snapshot requests without cancelling them, so late callbacks raced the replacements and piled up. It now generation-stamps requests, cancels on reset, backs off before retrying, and after two consecutive resets takes the frozen overlay down so a wedged web process reads as "controls faded out" instead of a hung UI (554f6e4). That hardens the recovery — it can't unstall your i915's snapshot path itself. If you still see the storms, two things would help: `journalctl -k | grep -iE "gpu hang|i915.*reset"` after a repro, and whether it correlates with anything else happening on the desktop at the time.

**Motion sometimes not revealing the UI** — found three holes, all closed: resizing while the chrome was hidden left the input-topmost overlay at its old size, so motion in the uncovered region never reached the page (0f826eb); the composite gate could desync from the chrome state in both directions, including a click-hide leaving a transparent snapshot loop running (b954df6); and a dead web process now reloads instead of going permanently silent (1a62cab).

**Hover highlight delay** — mostly by design, it turns out: the controls page has a deliberate 180ms CSS fade on hover states on *all* platforms; the Linux snapshot pipe adds maybe 30–80ms on top. I'm leaving it as is for now rather than diverging from the shared UI.

FeelThePoveR:
I did a retest on my Nvidia 4070 Super:
- **Flicker during drag action**
The flicker is still there, but greatly mitigated to the point I wouldn't consider this an issue (it only happens once every 1-2 secs basically

- **Double click issue**
I was also able to reproduce the double-click issue (sometimes needing to click twice for the action to apply) - this only happens when interacting through UI so for example spam clicking play/pause can make the action not work sometimes (the events are swallowed - nothing in the console), but spamming space to play/pause works without any issues (all actions result in a log being produced). 
This seems related to the updateTimer - raising the timer value makes it basically impossible to reproduce, lowering it makes it extremely easy to reproduce. 
Suprisingly enough this only happens on the play/pause button, I couldn't reproduce this on any other spammable action like changing of the aspect ratio or minimizing/maximizing

- **Dispose nitpicks**
The 2 nitpicky items here are basically irrelevant as they don't impact the functionality in any way
1. When closing the player during the playback both mpv and gtk throw errors and sometimes end in a SIGSEV
2. Very rarely dispose of the webkit UI (without closing the app) can throw an error (that is not impacting anything really). I got this twice in a half-hour timespan while really hammering the UI
`(java:78951): GLib-CRITICAL **: 19:21:08.622: Source ID 69 was not found when attempting to remove it`

Outside of those things I couldn't really find anything wrong on the Nvidia side

@skoruppa 
I also have a question for you - which settings should be applied on the desktop? 
I see that the desktop app still exposes Android settings like the "Playback Engine" for example so I wonder which settings should be handled on the desktop side. One of the settings I tried here was the preferred audio language and this doesn't get applied for example so I wonder if it's a gap in the linux bridge or just desktop behavior in general.

Paolo86cripple:
@ArjunKdaf could I help you test this on an ati/wayland setup?

ArjunKdaf:
@Paolo86cripple sure, any tests are welcome.

To build from source:

- deps: JDK 17, pkg-config, g++, and dev packages for libmpv, webkit2gtk-4.1, gtk3, libx11, libxcomposite, libxext
- clone https://github.com/ArjunKdaf/NuvioDesktop, branch `linux-desktop-support`
- create an empty `local.properties` in the repo root
- `./gradlew :composeApp:run`

A Wayland session is fine — the app runs through XWayland (the bridge pins X11), that's how it's been tested here too.

Worth checking: video plays, controls appear on mouse motion and fade out, cursor hides with them, resize and fullscreen (button, double-click, F11), Esc leaves the player, audio/subtitle panels, loading screen shows artwork instead of a bare spinner.

If something misbehaves, rerun with `NUVIO_BRIDGE_DEBUG=1` set and attach the log.

Minwalin:
@ArjunKdaf Could you update your Linux fork to Nuvio 0.1.16 when you have time? It looks like it’s still on 0.1.14, and I’d like to keep testing the latest version on Linux. Thanks!

ArjunKdaf:
@Minwalin done — the fork is on 0.1.16 now (merged current Dev, 144 commits). The merge was clean: upstream didn't touch the player stack, so the bridge needed no changes. Smoke-tested here on NixOS: playback, controls, fullscreen and the new tracking screens all behave.

@skoruppa @FeelThePoveR @Paolo86cripple if any of you can give it a fresh round on your setups, that would be welcome — same build steps as before, `NUVIO_BRIDGE_DEBUG=1` log if something misbehaves.

Minwalin:
> @Minwalin done — the fork is on 0.1.16 now (merged current Dev, 144 commits). The merge was clean: upstream didn't touch the player stack, so the bridge needed no changes. Smoke-tested here on NixOS: playback, controls, fullscreen and the new tracking screens all behave.

You're fast :)
 I updated to the latest linux-desktop-support branch, built the new 1.1.16 RPM, installed it successfully on Fedora, and I'm currently testing it. So far everything is working well. I'll let you know if I find any issues.
 
 UPDATE: I noticed the Tracking page shows Missing TRAKT_CLIENT_ID / TRAKT_CLIENT_SECRET and Missing SIMKL_CLIENT_ID. Is this expected for local builds? Should we provide our own API credentials in local.properties to test these integrations?
 
 Screenshoot: 
Screenshot_20260803_234035

FeelThePoveR:
@ArjunKdaf 
I did a quick retest retest again and the only issues I found are:
- ESC closes the player while in fullscreen - there was a [fix for this on the main branch](https://github.com/NuvioMedia/NuvioDesktop/commit/4abc3a73ddb3ec14f1f9d84473bc37f74f02138e) in the last week. The intended behavior is to exit fullscreen first and then exit the player
- sometimes needing to click multiple times to play/pause the player (I think it's updateTimer related)

From issues that I'm not sure are issues (I'm not sure about the Desktop behavior in general outside of Linux) - I would need somebody that has a Window or Mac setup to confirm (@Minwalin Maybe you could check those behaviors on Windows and confirm):
- Episode picker doesn't default to the current season
- The UI is getting hidden when episode/source/audio/subtitle picker is open
- Next episode prompt is also hidden if the player UI is hidden
- Preferred audio language settings don't translate to the player defaulting to the given language

Paolo86cripple:
Screenshot_20260805_112720
Screenshot_20260805_112929

Everything seems fine, except for issue described by @FeelThePoveR , and I also noticed that the  Clear Continue Watching option does not actually work. The continue watching cache is still there, unfortunately

Paolo86cripple:
Also, Nuvio cannot perform server sync.

ArjunKdaf:
@FeelThePoveR both issues run deeper than they look, thanks for the isolation work.

**Esc:** confirmed. Upstream's fullscreen-first fix lives in the controls page's keydown handler — which never gets keyboard focus on Linux (the overlay is composite-redirected), so my AWT key dispatcher was mapping Esc straight to "back". Fixed on the branch: Esc now exits fullscreen first, then leaves the player, same as `4abc3a7` intends.

**Multi-click on play/pause:** it's been around since the beginning and I was aware of it — the choice not to fix it yet was deliberate: the fix isn't in the bridge. Root cause, with your updateTimer hunch being exact: every 200ms state push re-renders the controls, and the play/pause icon's `<use href>` gets rewritten even when unchanged. WebKit rebuilds the icon's internal shadow tree on every such write; if a rebuild lands between mousedown and mouseup, the click is silently dropped — no event, no log. That's why space is immune (never enters the webview) and why only play/pause: its filled glyph covers the button center, while the aspect icon is never rewritten and the fullscreen glyph is hollow strokes. Since the mechanism sits in the shared controls page rather than the Linux bridge, other desktop platforms might be exposed to it as well — I haven't tested there, so I won't claim it.

The fix would be small (icon SVGs `pointer-events: none` so the stable button is always the click target, plus writing the icon `href` only on change) — but it lands in shared desktop files, and I'm keeping this PR strictly bridge-only. I'd rather submit that as a separate PR once the bridge is approved.

Two more found in my own testing, both outside the bridge, documenting rather than fixing for the same reason:

**Home Continue Watching can open an older episode.** On my setup, home showed a "continue" card for S04E08 while the series page correctly showed S04E10 mid-episode; clicking opened E08. Traced to the persisted next-up enrichment cache: it's only refreshed after the async server pull and is never invalidated when new progress is recorded, so right after launch the home row can serve a stale card, and the click opens the cached episode id without re-resolving. The likely fix is small (invalidate a series' cached card whenever progress is recorded for it) but it lives in shared state code — separate PR after the bridge lands, if wanted.

**Desktop doesn't apply the cache-size setting to TorrServer.** The app's P2P cache setting (default 2 GB) is applied on Android but the desktop engine never sends settings, so TorrServer runs its stock 64 MB in-RAM cache. Sequential playback is fine; resuming into a large file can thrash it (observed live: eviction loop while mpv builds the seek index, buffer counter cycling until the swarm outruns it). Same story — separate PR if wanted.

@Minwalin yes, expected for local builds — tracking keys (TRAKT/SIMKL client IDs) are injected into official builds; from-source builds need your own values in `local.properties` to test those integrations.

@Paolo86cripple both of yours are build-config, not bugs: server sync needs `NUVIO_SUPABASE_URL` / `NUVIO_SUPABASE_ANON_KEY` in `local.properties` (the same public values the official builds embed) — with an empty file it points at localhost. And "Clear Continue Watching" is a cache clear by design: it wipes the local artwork/metadata cache and re-pulls the list from the server, so the list itself comes back.

shscs911:
@ArjunKdaf Thanks a lot for working on this! I've been daily driving your build for the past week and everything just works out of the box except for the few bugs already reported by others. Apart from these, one minor bug I noticed was that the Linux build lacks the feedback toast shown when rewinding, fast-forwarding and increasing/decreasing volume. The keybindings work perfectly, but there's no visual indicators for these actions.
Screenshot 2026-08-06 234346
Screenshot 2026-08-06 234401
Screenshot 2026-08-06 234421

ArjunKdaf:
@shscs911 Confirmed. The toast is drawn by the controls page's own keydown handler — which never fired on Linux, because keys were mapped on the AWT side and sent straight to mpv, so the page never saw them. Fixed on my branch: the overlay now takes real X input focus, keys reach the page directly and its existing toast logic works unchanged, same as macOS/Windows. Verified on KDE and GNOME. Pushed — feel free to retest.

shscs911:
@ArjunKdaf Thanks a lot for fixing it. Action toasts work perfectly in my system now.

ofers:
Tested the current PR head `640d2180cd2a` on x86_64 Ubuntu 26.04, GNOME Wayland with XWayland, using OpenJDK 17 and Ubuntu's system mpv/libmpv 0.41.0.

Build interfaces reported by `pkg-config`:

- mpv client API: 2.5.0
- WebKitGTK: 2.52.3
- GTK: 3.24.52
- X11: 1.8.13
- XComposite: 0.4.6
- Xext: 1.3.4

Results:

- `:composeApp:buildLinuxPlayerBridge`: passed
- generated `libplayer_bridge.so`: x86-64 ELF, 34 JNI exports, no unresolved dynamic libraries
- `:composeApp:compileKotlinDesktop`: passed
- `:composeApp:desktopTest`: 102 suites / 637 tests / 0 failures / 0 errors / 0 skipped
- `:composeApp:run`: launched successfully with isolated XDG config/data/cache/state
- GNOME reported a focused, mapped, viewable `Nuvio` XWayland window (`WM_CLASS=com-nuvio-app-MainKt`) at 1280x783
- after the smoke test, the app process and X11 window both terminated cleanly; the worktree remained clean

The clean checkout had an empty `local.properties`, so startup logged a handled request failure against the default `https://localhost` Supabase endpoint. This did not crash the app, but it means this run validates Linux build/startup only, not configured catalogs, account services, or media playback.

I have not yet validated video embedding, controls, subtitles, hardware decoding, audio passthrough, or teardown during playback. I can test those next with legal local fixtures and report each capability separately.

ofers:
Packaging follow-up from the same PR head:

`./gradlew :composeApp:packageReleaseDeb` succeeds and produced `nuvio_1.1.16-1_amd64.deb` (142,444,700 bytes, SHA-256 `6f12ff394ac74287ebf59fd2c4420c4b12b4f4957f5379bf6265a8e7250bca18`).

However, the generated DEB does not declare the runtime dependencies of the Linux player bridge.

I extracted `native/linux/libplayer_bridge.so` from the packaged `composeApp-desktop` JAR. Its ELF `NEEDED` entries include:

- `libmpv.so.2`
- `libwebkit2gtk-4.1.so.0`
- `libjavascriptcoregtk-4.1.so.0`
- `libgtk-3.so.0` / `libgdk-3.so.0`
- `libcairo.so.2`
- `libgio-2.0.so.0` / `libgobject-2.0.so.0` / `libglib-2.0.so.0`
- `libXcomposite.so.1`
- `libX11.so.6`

The DEB's `Depends` includes X11 and general JVM/graphics libraries, but contains no mpv, WebKitGTK, GTK3, or XComposite package. The bridge is nested inside a JAR, so `jpackage` apparently does not discover its ELF dependencies automatically. The current `linux {}` packaging block sets only the icon.

This should be tested in a clean Ubuntu install/container before publishing the DEB. Possible directions are explicit distro package dependencies or deliberately bundling the required runtime libraries. Exact package names are distro/release-specific, so I have not proposed a list yet.

A smaller metadata issue: the generated control record says `Maintainer: Nuvio Media <Unknown>`.

If this packaging scope is welcome, I can prepare a focused fix plus a clean-environment verification test rather than mixing it into player behavior changes.

ofers:
Additional end-to-end Linux playback validation at PR head `640d2180cd2a8f574281a7a9301a55c847066aed`.

### Test method

I used a localhost-only Stremio-compatible addon serving FFmpeg-generated synthetic media. This avoided provider credentials and copyrighted media while exercising the real Nuvio path: catalog → metadata → stream discovery → embedded libmpv.

Environment remains x86_64 Ubuntu 26.04, GNOME Wayland/XWayland, OpenJDK 17, and system mpv/libmpv 0.41.0.

### Passed

- Fullscreen 3840×2160 launch and clean restart
- Addon installation and persistence
- Catalog, metadata, source selection, and HTTP Range playback
- H.264 720p/24 video and AAC stereo decode
- Pause/resume, timeline seek, mute/unmute, and Back/teardown
- Two AAC audio tracks enumerated and switched correctly
- Embedded English and Hebrew SubRip tracks enumerated
- Hebrew track label and RTL text rendered correctly/readably
- Subtitle Off and re-enable
- Profile, addon, catalog, and fullscreen state persisted across restart
- Leaving playback removed the mpv PipeWire stream while the app remained functional

### Audio-device finding

libmpv created an active 48 kHz stereo PipeWire stream, but it initially routed to the built-in analog sink, so there was no sound through the HDMI receiver. Moving only that live mpv stream to a temporary PipeWire sink backed by the machine's HDMI ALSA endpoint immediately restored AAC audio through the receiver.

The Linux bridge currently sets `audio-channels=auto` but does not expose or apply an `audio-device` choice/override. A user-selectable output or testable environment override, similar to the existing `NUVIO_MPV_GPU_CONTEXT` / `NUVIO_MPV_HWDEC` overrides, would make HDMI and passthrough validation practical without changing the system default sink.

### Subtitle behavior

The MKV marked English subtitles as default, but the bridge/UI called `selectSubtitleTrack off` when playback attached. Both embedded tracks were present and manual selection worked, including Hebrew. If disabling subtitles at every attachment is intentional, this is working as implemented; otherwise it prevents container-default subtitle selection.

### Cleanup diagnostics

Leaving playback worked, but I observed recoverable native cleanup warnings on some exits:

```text
GLib-CRITICAL: Source ID ... was not found when attempting to remove it
Gtk-CRITICAL: gtk_widget_get_window: assertion 'GTK_IS_WIDGET (widget)' failed
Gtk-CRITICAL: gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed
```

The player returned to details, audio stopped, and the application continued working. This may indicate duplicate/asynchronous GTK teardown worth guarding.

Not yet claimed/tested: confirmed VA-API use, HDR, refresh-rate switching, encoded-audio passthrough, or multichannel/lossless formats.

ArjunKdaf:
@ofers Thanks — this is thorough, and the synthetic-addon method is a clever way to exercise the real path without credentials.

**Packaging:** welcome — go ahead. I've kept my own changes to the bridge and Linux-gated paths and stayed away from shared behavior, but I won't say no to a collaboration. Since the Linux target only exists on my branch, open the PR against my fork's `linux-desktop-support` branch and it'll flow into this PR with you as author. Some findings to save you time: I checked the Compose plugin (1.11.1) — the DSL has `debMaintainer` (fixes the `<Unknown>`), but no hook for jpackage's `--linux-package-deps` and no raw-args passthrough (open request: JetBrains/compose-multiplatform#2236). So `Depends` needs post-processing of the packageDeb output: `dpkg-deb -R`, append, `dpkg-deb -b` — dpkg-deb is guaranteed present since jpackage itself shells out to it. On package names: `libmpv2, libwebkit2gtk-4.1-0, libxcomposite1, libxext6` sidesteps the t64 renames entirely (webkit transitively pulls javascriptcore/gtk3/cairo/glib/x11; note `libXext.so.6` is in the NEEDED set too). Ubuntu 22.04 is out of reach for the packaged DEB: its repos stop at libmpv1, and the bridge links libmpv.so.2 — jammy users would need a newer mpv from a PPA or source build, which a package dependency can't express.

**Audio device:** not a misroute — the bridge never sets `ao` or `audio-device`, so mpv's PipeWire stream goes to your system default sink, which is standard behavior (a stream move in pavucontrol is remembered by WirePlumber). The override was cheap though, so I've added `NUVIO_MPV_AUDIO_DEVICE` next to the existing `NUVIO_MPV_GPU_CONTEXT`/`NUVIO_MPV_HWDEC` overrides — values from `mpv --audio-device=help`. A user-facing output picker would be new shared UI across every desktop platform (none has one today) — bigger than a quick fix, but fair game as a follow-up if someone wants to take it on.

**Subtitles:** intentional, and identical on macOS/Windows. Subtitle choice is owned by the app, not the container: with the default "Preferred subtitle language: None", shared code explicitly deselects whatever mpv auto-picked from the container, and manual per-title choice persists. All three bridges just receive the same disable call. Whether "None" should mean "leave the container default alone" instead of "force off" is a product question — that behavior is shared across all platforms, so I'd want the maintainers' word before anyone changes it.

**Teardown warnings:** confirmed real — thanks. Root cause of the widget asserts: player close races Compose's canvas teardown against the native dispose; when the canvas X window dies first, the cascade destroys the overlay at the X level and GTK finalizes the overlay toplevel itself — the bridge's teardown then touched a freed widget. Fixed on my branch, bridge-only: the bridge now tracks external destruction of the overlay, teardown skips what GTK already destroyed, source removal is guarded, the script-message handler and webview setup got liveness checks, and dispose is idempotent. Verified over repeated closes: the `gtk_widget_get_window`/`gtk_widget_destroy` asserts are gone. The `Source ID` warning I never managed to reproduce, but its mechanism (removing a source that had already removed itself during the close race) is guarded now too. Remaining noise: a few GLib/GDK frame-clock criticals can still appear at close — they're emitted inside GTK's own forced destroy of the externally-killed window, before any bridge code runs. The full cure is reordering the shared desktop teardown so the player is disposed before the canvas goes away — that touches macOS/Windows too, so it's a candidate follow-up rather than something I'll fold in here. They're cosmetic; the app recovers fully. Feel free to retest exits once this is pushed.

ofers:
Packaging follow-up is ready in the requested fork workflow:

- https://github.com/ArjunKdaf/NuvioDesktop/pull/2
- target: `ArjunKdaf/NuvioDesktop:linux-desktop-support`
- author: `ofers`

It sets the generated DEB maintainer to `Nuvio Media <hello@nuvio.tv>` and adds the approved runtime relationships: `libmpv2`, `libwebkit2gtk-4.1-0`, `libxcomposite1`, and `libxext6`.

Validation completed on Ubuntu x86_64/JDK 17:

- normal and release DEBs built and passed artifact verification;
- disposable fixtures covered multiline/versioned/alternative relationships, repeated patching, paths with spaces, corrupt data archives, and ownership rejection;
- 637 desktop tests passed with no failures, errors, or skips;
- the release DEB completed a clean APT install, packaged-bridge linkage check, visible installed-app launch, and full purge;
- an independent staged-diff review found no blocking correctness or security issues.

The separate `NUVIO_MPV_AUDIO_DEVICE` Yamaha-route and repeated player teardown retest is still pending because I am not currently at the TV; I am not treating that behavior as validated yet.

ofers:
TV retest update for the Linux audio override and repeated player teardown:

### Test setup

- route: ThinkPad HDMI → Yamaha RX-A2A → TCL TV
- process-local override: `NUVIO_MPV_AUDIO_DEVICE=alsa/hdmi:CARD=PCH,DEV=0`
- isolated Nuvio profile
- localhost-only synthetic 60 s H.264 + stereo AAC fixture
- global PipeWire default remained the host's analog sink

### Result

The override works on first playback: video played and AAC audio was clearly audible through the Yamaha. During playback, Nuvio owned `/dev/snd/pcmC0D3p`, confirming the direct HDMI route.

I also reproduced a teardown/re-entry issue:

1. First playback: video + Yamaha audio.
2. Back returned cleanly to the catalog, but the HDMI PCM remained owned by Nuvio after >3 s; no native-player disposal line appeared.
3. Second playback created a new native handle and rendered video, but there was no audible HDMI audio.
4. Back again returned cleanly; the HDMI PCM remained held.
5. Third playback also rendered video without audio.
6. Full app termination released the PCM. That shutdown emitted GTK/GLib warnings (`invalid (NULL) pointer`, `g_signal_handler_disconnect` assertion, and `gdk_frame_clock_end_updating` assertion).
7. Relaunching with the same profile and override restored Yamaha audio on the first playback.
8. On that restarted run, Back released the PCM correctly and shutdown did not repeat the warnings.

This points to a nondeterministic player teardown/resource-release race: when the old instance retains the direct ALSA device, subsequent instances have video but no audio until the app restarts.

Scope limitation: this validates decoded AAC/PCM routing only. It does **not** validate compressed passthrough, HDR, refresh switching, or hardware decoding.

Cleanup completed: test package purged, Nuvio process absent, HDMI PCM released, and the global audio default remained unchanged.

ofers:
Clarification on step 6 above: the process was terminated by the test harness after returning to the catalog, rather than through a user-initiated normal Quit action. The GTK/GLib warnings therefore document that process-boundary shutdown path and should not be read as a confirmed reproduction of the normal UI Quit path. The Back/re-entry audio failure and retained PCM observations are unaffected.

FeelThePoveR:
@ArjunKdaf 
The play/pause seems to have been fixed on the main branch here (https://github.com/NuvioMedia/NuvioDesktop/commit/a938caf5f73c5625f10d1763128ad24cfb699767), can we rebase the branch and get this fix in here as well?

@skoruppa @tapframe 
Can we get some attention on this PR? 
From the testing done by the community this should be pretty done and I imagine that having this on main would make more sense than pinging the author to rebase and add the fixes from main every once in a while (like the ESC fix or the play/pause fix).

ofers:
Focused native-player teardown follow-up is ready for review:

- https://github.com/ArjunKdaf/NuvioDesktop/pull/3
- commit: `6c445999773629cfda69b0586b82871f4d170f81`
- desktop suite: **105 suites / 666 tests / 0 failures / 0 errors / 0 skipped**
- final independent exact-tree review: `passed=true`, with no security concerns, logic errors, or test gaps
- exact serial DEB SHA-256: `604c831ba1c0df3311ffef2412f99ab636a05e65805dcd4789ddac92b9b1cec8`
- exact rebuilt/installed DEB: **2/2** Xvfb playback → Back cycles acquired, released, and reacquired the real direct-ALSA HDMI endpoint; Nuvio remained alive and the global sink remained analog
- physical ThinkPad → Yamaha RX-A2A → TCL validation on the predecessor reviewed build: **5/5** visible-video/audible-Yamaha cycles returned cleanly and released PCM ownership

The review found and closed an additional queued-EDT-disposal race before submission. The regression was RED before the ordering barrier and GREEN afterward; a separate watchdog test proves timeout cannot later invoke navigation success.

Current upstream `Dev` is `21b5c3d8` and includes the separate play/pause fix `a938caf5`. A read-only `git merge-tree` simulation reports the same two pre-existing #295 conflicts (`Main.kt` and `DesktopAppFullscreen.kt`) both with and without the teardown commit. The teardown changes auto-merge in the shared player files and add no reported conflict.

Scope remains teardown/resource release only; this does not claim passthrough, HDR, refresh switching, Dolby Vision, CEC, or eARC validation.

kali113:
Really want to see this PR land! There are plenty of people volunteering to test (myself included), but a few things need attention before maintainers can review:

### Blockers
* **Rebase on upstream `dev`:** The branch has fallen behind and has merge conflicts. Note that the play/pause issue flagged by @FeelThePoveR (fixed upstream by @Laskco in a938caf).has already landed upstream. 
* **Pick up the HDMI fix:** It's worth pulling in @ofers's fix ([#3](https://github.com/ArjunKdaf/NuvioDesktop/pull/3)) for the TV HDMI audio bug before rebasing.
* **PR Template:** Please fill out the standard PR template sections and link issue #27 so the CI bot doesn't flag it.
* **Direction Confirmation:** @tapframe mentioned Linux should use native mpv + webview (instead of copying frames into Compose). Can @tapframe confirm this implementation aligns with that direction?

### Scope & Testing
* **Keep the scope focused:** Skip Flatpak, AppImage, AUR, and native Wayland for this PR. AWT relies on X11, and XWayland is completely fine for an alpha. Keeping the scope tight avoids stalling.
* **Smoke test:** A quick run across both Mesa and NVIDIA would help verify basic stability once rebased.

### Nice-to-Haves (Optional / Follow-up)
* A minimal Ubuntu CI job to build the bridge and desktop test.
* A short "Running on Linux" note covering dependencies, bridge debugging, and NVIDIA quirks.

Once the rebase and template are sorted, this should be ready for maintainer review!

kali113:
Tested this on Arch Linux (`GNOME Wayland` + `XWayland`, `Mesa NV106 / nouveau`, `JDK 21`, `mpv 0.41`):

It builds and runs cleanly. The sign-in screen renders fine, and playing a local 20s H.264+AAC clip through `libmpv` worked as expected (`x11egl` initialized, PipeWire audio worked, and it played through to EOF).

A couple of quick observations:
* **Smoke test crash:** `NUVIO_DESKTOP_SMOKE_PLAYER_URL` crashes out of the box with `Platform density is unavailable outside NuvioTheme`. Wrapping the smoke test path inside `NuvioTheme` resolved it and allowed playback to proceed.
* **Hardware decode fallback:** On this GPU/driver combo (`nouveau`), `hwdec` didn't engage and dropped back to software decoding. Video still rendered properly without breaking—just flagging for context.

Happy to re-test whenever this gets rebased onto `dev`!

skoruppa:
@ArjunKdaf yeah, I think that after resolving conflicts and addressing what @kali113 wrote about this can be merged as experimental support 

@KhooLy want to add something?

@tapframe any objections or comments? :P

posilash:
Tested on NVIDIA proprietary, rebased onto current `Dev`.

**Environment**

- NVIDIA GeForce RTX 3070 Laptop GPU, proprietary driver 610.57.04, hybrid with an AMD Vega integrated GPU
- Arch Linux, kernel 7.1.8, Hyprland (Wayland session, so the player runs on XWayland)
- OpenJDK 26.0.2.1, mpv 0.41.0, libplacebo 7.360.1

**Result: builds and plays cleanly.** Rebasing onto current `Dev` took three conflicts (`composeApp/build.gradle.kts`, `Main.kt`,`DesktopAppFullscreen.kt`), all small. 4K HEVC HDR played embedded in the window for about 40 seconds with 7.1 audio through PipeWire, no dropped frames, clean EOF.

**`x11egl` fails on the proprietary driver, and the `x11vk` fallback saves it.** This complements @kali113's nouveau run, where `x11egl` initialized fine:

```
mpv_initialize failed (gpu-context=x11egl): invalid parameter
vo/gpu: Failed initializing any suitable GPU context!
cplayer: Error opening/initializing the VO window.
gpu-context x11egl failed to initialize; using x11vk instead
mpv initialized OK
```

So the two-attempt logic is load bearing on NVIDIA, and worth keeping.

**Hardware decoding works here.** Where nouveau fell back to software, this engages:

```
vd: Using hardware decoding (nvdec-copy).
vd: Decoder format: 3840x2160 p010 bt.2020-ncl/bt.2020/pq/limited
cplayer: VO: [gpu] 3840x2160 p010
```

The `x11vk` plus `nvdec-copy` pairing behaves as intended.

**Two minor observations**
1. mpv 0.41 logs a deprecation: `option --vd-lavc-software-fallback was
   replaced with --hwdec-software-fallback and might be removed in the
   future`, once per init attempt.
2. On exit: `g_signal_handler_disconnect: assertion 'G_TYPE_CHECK_INSTANCE
   (instance)' failed` and `gdk_frame_clock_end_updating: assertion
   'GDK_IS_FRAME_CLOCK (frame_clock)' failed`. Cosmetic, after teardown starts.

Separately, one thing worth knowing for anyone running this on HDR content:
the controls are pushed to mpv through `overlay-add`, and under`vo=gpu-next` mpv reinterprets that bitmap in the video's colorspace, so the interface renders wrong on HDR files. `vo=gpu` is unaffected. I filed mpv-player/mpv#18392 for it. Not an issue for this PR, since it forces `vo=gpu`.

I have the rebase onto current `Dev` done and building if that is useful.

**What I have in my fork**, at https://github.com/posilash/NuvioDesktop/tree/linux:
1. This branch merged onto current `Dev` (0.1.20), with the three conflicts above resolved. Builds and runs, verified above. This is the rebase @kali113 listed as the first blocker, so please take it if it saves time.
2. A fix for a signature drift this branch has picked up: the Kotlin declaration of `applySubtitleStyle` passes ten arguments while the Linux JNI entry point takes eight, so the libass toggle and the SDH stripping preference never reach mpv. Both are applied now, matching the Windows bridge.
3. Beyond the scope @kali113 asked to keep tight, and only mentioned so it is clear what is and is not part of the above: a desktop entry with the `nuvio://` and `stremio://` scheme handlers the desktop deep link path already parses, an AUR PKGBUILD, and a rework of how the bridge applies mpv options so a user config can be layered under Nuvio's own settings.
Happy to raise any of those separately once this lands.

ArjunKdaf:
Branch is now current with `Dev` and the checklist from @kali113 is complete:

1. **Rebased onto current `Dev` (0.1.20)** — same three conflicts @posilash hit (`build.gradle.kts`, `Main.kt`, `DesktopAppFullscreen.kt`), resolved the same way: my GTK early-init runs before upstream's renderer/Sentry/QuickJS setup, and the fullscreen file keeps upstream's new toggle signature plus the Linux fullscreen enforcement (the mutter exit bug upstream fixed for macOS only).
2. **@ofers' teardown PR is merged** — the release barrier that disposes the native player before navigation, closing the held-PCM-on-Back bug from his TV test.
3. **@posilash's `applySubtitleStyle` signature fix is cherry-picked, authorship kept** — the SDH stripping and libass preferences that landed in #381 now actually reach mpv on Linux. Thanks for catching that. Good data on the NVIDIA side too — the `x11egl`→`x11vk` fallback engaging on proprietary, `nvdec-copy` hardware decoding, and a Hyprland/XWayland session all confirmed; that lines up with @FeelThePoveR's proprietary runs going back to the first builds and @kali113's nouveau/`x11egl` path.
4. **The smoke-harness crash @kali113 found is fixed** (the harness now runs inside `NuvioTheme`).
5. **PR template filled**, linked to #27.

Test suite on the merged branch: 123 suites / 744 tests / 0 failures. Bridge and the new vendored compose-media-player native build both compile.

One packaging note: upstream's vendored trailer player brings GStreamer into the desktop build, so the DEB runtime dependencies will need `gstreamer` and `gst-plugins-base` alongside the existing mpv/WebKitGTK set — @ofers, flagging it since that's your area.

@FeelThePoveR — you've been testing this since the first builds; the branch now carries everything from current `Dev`, so if you want one more pass on the 4070 before this goes in, now is the time.

@skoruppa ready from my side.

ArjunKdaf:
Exercised the new vendored trailer player (compose-media-player) on Linux for the first time — it had never run on this platform. Trailers now play in-app; two fixes pushed, both confined to the library's Linux backend (no Windows/macOS effect):

1. **Whole-app freeze on every trailer click** — `LinuxVideoPlayerState.clearError` does `runBlocking { withContext(Dispatchers.Main) }`; called from a Compose effect on the main thread, `runBlocking` parks the EDT and the dispatch queues onto that same parked thread. Windows never reaches this code (trailer cards open externally there) and macOS has its own backend, which is why it went unseen. Fixed with `Dispatchers.Main.immediate`. Probably worth sending upstream to kdroidfilter/compose-media-player at some point — any Linux consumer of the library hits it.
2. **googlevideo rejects GStreamer's default user agent** (403 → typefind "not enough data") — the native player now sets a browser UA via playbin's `source-setup`, matching what the extraction probe sends.

@ofers packaging note, refining the earlier GStreamer one: the runtime deps for trailers are `gstreamer1.0-plugins-good` (MP4 demux), `gstreamer1.0-libav` (H.264/AAC decode) and `glib-networking` (TLS for libsoup — without it every https stream fails with "TLS support is not available"). Most desktop installs have them; the DEB Depends should list them anyway.

FeelThePoveR:
@ArjunKdaf 
I recently had an opportunity to leave the Nvidia card behind and get a new one for cheap so now I'm testing using a Radeon 9070XT on Fedora 44 so I guess posilash's testing of the build needs to be sufficient unless someone else with Nvidia proprietary shows up.

Anyways from my testing the player works perfectly fine, the only nitpick remaining would be that the commit I linked that fixed the play/pause bug (https://github.com/NuvioMedia/NuvioDesktop/commit/a938caf5f73c5625f10d1763128ad24cfb699767) also had a behavioral change in the Windows/MacOS bridges - codewise at least it made it so that pressing play at the end of playback would seek to the start of the playback, this is missing from the Linux bridge.

Outside of that I didn't find any more issues so I would say the PR looks good to me!

KhooLy:
Hi, thanks for the PR. Tap wants to know if you are willing to continue help if things for linux goes wrong after release like maybe bugs occur etc.

ArjunKdaf:
Hi, yes — happy to continue helping with Linux bugs and fixes after release. One heads-up: in a month or two I'll be leaving on a trip for around 5–6 months, with little to no internet access, so I'll be effectively unreachable during that time. Until then, count me in.

ofers:
Thanks — I checked the new Linux comments against the current `linux-desktop-support` head.

- The generated `1.1.20` DEB did omit `gstreamer1.0-plugins-good`, `gstreamer1.0-libav`, and `glib-networking`. I opened a focused packaging fix in https://github.com/ArjunKdaf/NuvioDesktop/pull/5. The rebuilt exact artifact contains all three and passes `scripts/verify-linux-deb.sh`.
- The original native teardown PR was merged before its owner-review remediation landed. The complete remediation is now rebased onto the current Linux head in https://github.com/ArjunKdaf/NuvioDesktop/pull/4. Current-base verification is 124 suites / 754 tests / 0 failures plus an exact-DEB direct-HDMI release-before-navigation smoke.
- The Linux replay-at-EOF parity observation is valid, but it is a separate playback-behavior change and is intentionally not bundled into either focused PR.

Both follow-up PR template checks pass. I have not merged or closed either PR.

ArjunKdaf:
Status on the branch since the last update.

Two contributor PRs landed. The first hardens the player teardown path: release-before-navigation now contains failures at every boundary, and each attempt has a single owner, so a late or duplicate native callback can't act on a retry or navigate twice. The second fixes the trailer player not replaying after end of stream on Linux — end of playback is now driven by authoritative GStreamer EOS rather than a duration heuristic.

The teardown work also caught a regression I had missed. On desktop the player surface stayed mounted through every source gap rather than only during teardown, so switching stream or hitting next episode left the window black for the whole re-resolution instead of showing the loading overlay. Retention is now scoped to an active back release, and ordinary source gaps behave the same as on the other platforms. Confirmed here.

Full desktop suite 757/757, media player suite 82/82, both on Linux. The branch is still conflict-free against Dev.

manavchaudhary1:
Horizontal scrolling is not working on Linux desktop.

- Episode lists on series/details pages
- Continue Watching
- Other horizontal carousels/lists on the Home page

Two-finger horizontal touchpad scrolling does not move these lists, and Left/Right arrow-key navigation also does not work.

Didn't try on any other OS.
& build rpm package using PR #296

Issue #481, opened 2026-08-25
[Bug]: Previously selected resize mode not applied at stream start


ThEgIGN:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.1.20-alpha

### Install method

GitHub Release desktop installer/package

### Platform

Windows (desktop)

### Desktop hardware

Windows 11 laptop with Intel i7-14700HX and NVIDIA RTX 4060 Laptop GPU

### OS version

Windows 11 Pro 25H2

### Area (tag)

Playback (start/stop/buffering)

### Bug description

Every time I start stream, UI controls remember previous resize mode and show it correctly, however player doesn't remember it. So, every time stream starts, controls correctly show (Fit/Zoom/Stretch) mode from previous session, but player always plays video in Fit mode.

### Steps to reproduce

1. Open stream
2. Change resize mode to anything but Fit (Zoom or Stretch)
3. Close stream
4. Open either same stream or another one

### Expected behavior

Player correctly displays video in last resize mode that user selected

### Actual behavior

Player displays video in Fit mode

### Frequency

Always

### Did this work before?

No, it never worked

### Logs (required for crash reports)

```shell

```

PR #482, opened 2026-08-25
Fix resize mode not being applied when stream starts

ThEgIGN:
## Summary

- Call `setResizeMode` directly from function `createPlayer` (it will only be called once per stream and when player is created)
- Grab resize mode from `controlsState`
- If `controlsState` is corrupted in any way, set resize mode to default value Fit

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

Controls always keep value of previous resize mode that user selected, so player should set its mode accordingly.

## Desktop scope

This change is related to shared desktop code.

## Issue or approval

Fixes issue [#481](https://github.com/NuvioMedia/NuvioDesktop/issues/481)

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Only `createPlayer` in `NativePlayerController` is changed. There might be a way to grab `PlayerResizeMode` value from storage, similarly like `applyRememberedVolume` does it. Except `applyRememberedVolume` has its own separate storage.
Another way this could be done is when `PlayerEngine` calls `setResizeMode` for the first time with value from storage and that call gets ignored because player doesn't exist yet, specific mode could be saved in random value in `NativePlayerController`, which would then be accessed only once by `setResizeMode` in `createPlayer`.

## Testing

Tested by changing resize mode to either Zoom or Stretch. Then I tried:

- Closing and opening same stream
- Closing and opening different source for same episode
- Closing and opening different series
- Changing source mid stream
- Changing episode mid stream
- Closing and opening application before opening stream

Tested with basic `println()` and confirmed that `setResizeMode` in `createPlayer` gets called only once when stream starts.

## Screenshots / Video

Not a UI change

## Breaking changes

None

## Linked issues

Fixes issue [#481](https://github.com/NuvioMedia/NuvioDesktop/issues/481)

### DIFF
--- a/composeApp/src/desktopMain/kotlin/com/nuvio/app/features/player/desktop/NativePlayerController.kt
+++ b/composeApp/src/desktopMain/kotlin/com/nuvio/app/features/player/desktop/NativePlayerController.kt
@@ -65,6 +65,9 @@ internal class NativePlayerController(
 
         @Volatile
         var rememberedVolumeLevel: Float = DesktopPlayerVolumeStorage.loadVolumeLevel() ?: 1f
+
+        @Volatile
+        var rememberedResizeMode: PlayerResizeMode = PlayerResizeMode.Fit
     }
 
     private data class ReleaseCallback(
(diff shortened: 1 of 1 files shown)


ThEgIGN:
Realized languages other than English exist, so I fixed hard coded solution in second commit. Now there is `rememberedResizeMode` variable that has default Fit value that changes correspondingly. `setResizeMode` in `createPlayer` still gets called only once.

Reason why `rememberedResizeMode` is set every call of `setResizeMode` is because originally I wanted to only set it when handle doesn't exist, but problem occurs when I switch stream source during playback. In that moment handle exists and `rememberedResizeMode` doesn't get set to correct value.

Tested it again the same way mentioned above, but this time I also changed languages multiple times and combined previous tests during those switches.

PR #493, opened 2026-08-26
ci(linux): Add Deb, Rpm, AppImage and Flatpak github workflows 

### LINKED ISSUE #324: [Feature]: Adopt Flatpak to simplify Linux maintenance and distro support
### Area (tag)

### Problem statement

Right now, I notice that for the Linux build, is being planned as a .deb installer. While that works well for Ubuntu and Debian users, it creates an unnecessary barrier for anyone running other popular Linux distributions like Fedora, Arch, or openSUSE.

For the Nuvio team, supporting a .deb package directly means taking on a lot of extra troubleshooting, Because every Linux distribution handles system files and software libraries slightly differently, users on different setups will inevitably run into unique bugs. Instead of focusing on building features and fixing core bugs, maintainers end up spending time diagnosing issues caused by a user's specific system configuration.

### Proposed solution

Switching to Flatpak as the primary Linux format solves both sides of this equation.

Flatpak packages the application alongside its required dependencies in a self-contained bundle. For maintainers, this means you only have to build, test, and release a single file rather than managing multiple Linux formats. For users, it guarantees that NuvioDesktop runs in the exact same reliable environment on every single computer, regardless of which Linux flavor or version they happen to be using.

No, this is only a proposal

AppImage: Easy to run anywhere, but doesn't handle desktop integration or background updates as smoothly for everyday users.

Snap: Cross-platform, but less popular among non-Ubuntu users and harder to configure for general distribution.

Making Flatpak the main release target gives Nuvio Desktop the widest possible reach with the lowest ongoing maintenance effort.

FeelThePoveR:
## Summary

Adding workflows for Linux deb, rpm, appimage and flatpak builds

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [X] Small maintenance only, with no UI or behavior change
- [ ] Docs accuracy fix

## Why

This change is needed to make the newly introduced Linux support buildable using github actions.

## Desktop scope

Platforms: Linux

## Issue or approval

Fixes: https://github.com/NuvioMedia/NuvioDesktop/issues/324

## UI / behavior impact

- [X] No UI change
- [X] No behavior change

## Policy check

- [X] I have read and understood `CONTRIBUTING.md`.
- [X] This PR is small, focused, and limited to one problem.
- [X] This PR is scoped to the desktop app, desktop packaging, desktop documentation, or shared code required for desktop behavior.
- [X] This PR is not cosmetic-only.
- [X] Any UI change fixes a linked glitch/bug and includes visual proof, or this PR has no UI change.
- [X] Any behavior change fixes a linked bug/regression or has explicit approval, or this PR has no behavior change.
- [X] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [X] This PR does not add dependencies, architecture changes, migrations, or product-direction changes without explicit approval.
- [X] I listed the testing performed below.

## Scope boundaries

This change only introduces new linux workflows, linux package patch scripts and build targets.

## Testing

RPM, Flatpak and AppImage tested on Fedora 44 using RX 9070XT GPU. RPM and DEB were also tested using distrobox to ensure that the package install also fetches the necessary dependencies and that the app runs well when installed on a clean system.

## Screenshots / Video

Not a UI change

## Breaking changes

None

## Linked issues

Issue: https://github.com/NuvioMedia/NuvioDesktop/issues/324


KhooLy:
Looks good overall. What do you think about pinning appimagetool to a stable version + SHA256, and removing || true from the required GNOME Platform/SDK install? The VAAPI extensions can remain optional.

FeelThePoveR:
@KhooLy Both changes are done. Pinning AppImageTool against 1.9.1 is fine, the continuous build only contained a fix for when the tool is being run via a symlink so nothing that concerns this job.

PR #496, opened 2026-08-27
fix(linux): make horizontal wheel and touchpad scrolling work

### LINKED ISSUE #495: [Bug]: Horizontal scrolling does not work on Linux and scrolling left navigates back
### Pre-flight checks

### App version / commit

`Dev` at `24f63e42`

### Install method

Built from source

### Platform

Linux (desktop)

### Desktop hardware

12th Gen Intel Core i7-12650H, Intel UHD Graphics + NVIDIA RTX 3070 Mobile

### OS version

NixOS 25.11, kernel 6.12.93, KDE Plasma on Wayland (the app runs under Xwayland)

### Area (tag)

Keyboard / Mouse input

### Bug description

Horizontal scrolling does nothing on Linux, and scrolling to the left navigates the app backwards instead.

Two-finger horizontal touchpad scrolling, and the tilt wheel on a mouse, do not move the episode list, Continue Watching, or any horizontal row on Home. A left tick also takes me back one screen, so scrolling sideways on a details page returns me to where I came from.

The cause is button numbering. X11 delivers the horizontal wheel as buttons 6 and 7, and Java's XToolkit reports those as AWT buttons 4 and 5. Compose derives its `PointerButton` from the AWT button number minus one, so button 4 lands on `PointerButton.Back` and button 5 on `PointerButton.Forward`. The ticks never reach the scroll path at all: `ComposeSceneMediator` only produces a horizontal scroll from a `MouseWheelEvent` with shift held (`shift ? Offset(rotation, 0) : Offset(0, rotation)`), and the horizontal wheel never becomes a `MouseWheelEvent` on X11.

This is the same numbering quirk behind #294 and the thumb buttons, seen from the other end.

### Steps to reproduce

1. Run the desktop app on Linux.
2. Open Home, or a series' details page with the episode list in Horizontal card mode.
3. Put the pointer over a horizontal row and scroll sideways with two fingers on a touchpad, or with a tilt wheel.
4. Scroll left, then scroll right.

### Expected behavior

The row scrolls sideways, the way Shift + wheel already does, and no navigation happens.

### Actual behavior

The row does not move in either direction. Scrolling left navigates back one screen. Scrolling right does nothing at all.

### Frequency

Always

### Did this work before?

No, it never worked

Not applicable — nothing fails and nothing is logged; the input is simply routed to the wrong handler.

### Anything else? (optional)

Reported on the Linux support PR after it merged: https://github.com/NuvioMedia/NuvioDesktop/pull/295#issuecomment-5421204384

Shift + vertical wheel scrolls these rows correctly on Linux, which confirms the scroll path itself is fine and only the event shape is wrong.

Windows and macOS are unaffected: their toolkits number these buttons the way Compose reads them.

I have a fix and will open a PR against `Dev`.

ArjunKdaf:
## Summary

- Translate the X11 horizontal wheel into the shift-modified `MouseWheelEvent` Compose reads as a horizontal scroll.
- Stop a horizontal tick from reaching the app as `PointerButton.Back`/`Forward`, which navigated instead of scrolling.
- Move the existing thumb-button remap onto the same pushed `EventQueue`, since an `AWTEventListener` cannot swallow the event it replaces.

## PR type

- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

X11 delivers the horizontal wheel as buttons 6 and 7, and Java's XToolkit reports those as AWT buttons 4 and 5. Compose derives its `PointerButton` from the AWT button number minus one, so on Linux a horizontal tick arrives as `PointerButton.Back` or `PointerButton.Forward`. Scrolling a row sideways therefore moved nothing and, to the left, navigated the app back one screen.

The scroll path itself is fine — Shift + wheel already scrolls these rows on Linux. `ComposeSceneMediator` builds `Offset(rotation, 0)` when shift is held and `Offset(0, rotation)` otherwise, so a horizontal scroll only exists as a `MouseWheelEvent` with shift, and the horizontal wheel never becomes a `MouseWheelEvent` on X11. The fix is to build that event from the button press.

The press also has to be swallowed, or the back navigation stays. An `AWTEventListener` can only observe, so the translation moves to a pushed `EventQueue`, and the thumb-button remap added in #473 moves with it rather than living in a second mechanism.

## Desktop scope

Linux only. `LinuxExtendedMouseButtons.kt` is desktop code that returns immediately unless `os.name` is Linux, and the file is the only one touched. Windows and macOS toolkits already number these buttons the way Compose reads them and never reach the translation. No shared or common code is involved.

## Issue or approval

Fixes #495.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

A horizontal wheel tick now scrolls the row under the pointer instead of navigating backwards.

## Policy check

## Scope boundaries

The thumb back button keeps exactly the behaviour it has today: AWT button 6 is dispatched as button 4 so `PointerButton.Back` matches. `PointerButton.Forward` is still deliberately unmapped, because nothing in the app acts on it.

Not changed: the vertical wheel, `controls.js` and the in-player controls, the native player bridge, scroll speed or smooth-scrolling behaviour anywhere, and any shared or non-Linux code. No dependencies added.

## Testing

Built from source on NixOS 25.11 (kernel 6.12.93, KDE Plasma on Wayland, the app under Xwayland), on `Dev` at `24f63e42`.

Desktop test suite: 133 suites, 790 tests, 0 failures, 0 skipped.

Manual, on a laptop touchpad with two-finger horizontal scrolling:

- Home carousels, Continue Watching, and a series episode list in Horizontal card mode all scroll sideways now, in both directions, with the row moving the way the fingers move.
- Scrolling left no longer navigates back.
- Shift + vertical wheel still scrolls these rows, unchanged.
- Thumb back button still navigates back outside the player, and inside the player back and forward each still seek 10s per press through the native bridge, with no double step.
- Left and right clicks, drags on the seek and volume sliders, and vertical scrolling are all unchanged. Middle click does nothing, as before this change — there is no handler for it in the app.
- The window still closes normally from its own button, which is the failure mode to watch for when touching AWT startup on Linux.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #495.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/com/nuvio/app/LinuxExtendedMouseButtons.kt
+++ b/composeApp/src/desktopMain/kotlin/com/nuvio/app/LinuxExtendedMouseButtons.kt
@@ -1,30 +1,55 @@
 package com.nuvio.app
 
 import java.awt.AWTEvent
+import java.awt.Component
+import java.awt.EventQueue
 import java.awt.Toolkit
+import java.awt.event.InputEvent
 import java.awt.event.MouseEvent
+import java.awt.event.MouseWheelEvent
 
-// X11 numbers the mouse thumb buttons 8 and 9. Java's XToolkit reports the
-// vertical wheel as MouseWheelEvent and keeps the horizontal wheel as buttons 4
-// and 5, so the thumb buttons arrive as AWT buttons 6 and 7. Compose derives
-// PointerButton.Back from the 4th button, so the back navigation in
-// MainAppContent never matched and the button did nothing on Linux.
+// X11 numbers the buttons above the wheel differently from the other desktop
+// toolkits and Java's XToolkit passes that numbering straight through. The
+// vertical wheel becomes a MouseWheelEvent, but the horizontal wheel stays a
+// pair of button presses -- X11 buttons 6 and 7, reported as AWT buttons 4 and
+// 5 -- so the thumb buttons (X11 8 and 9) arrive as AWT buttons 6 and 7.
 //
-// Re-post the press and release under the button number Compose expects. The
-// shared code stays untouched and keeps working unchanged on Windows and macOS,
-// where the toolkit already numbers these buttons the way Compose reads them.
+// Compose derives its PointerButton from the AWT button number minus one, which
+// leaves two separate defects on Linux:
 //
-// Only Back is mapped. Nothing acts on PointerButton.Forward: there is no
-// forward navigation anywhere in the app, and the player's own seek arrives
-// through the native bridge rather than through AWT, so mapping the second
-// thumb button would post events that no handler consumes.
+//   * the thumb back button arrives as button 6, never matches
+//     PointerButton.Back, and back navigation does nothing;
+//   * a horizontal wheel tick arrives as button 4 or 5, which is exactly what
+//     Compose reads as Back and Forward, so scrolling a row sideways navigates
+//     the app backwards instead of moving the row.
+//
+// A horizontal scroll only reaches Compose as a MouseWheelEvent with shift held
+// -- ComposeSceneMediator builds Offset(rotation, 0) when isShiftDown and
+// Offset(0, rotation) otherwise -- so the wheel buttons have to be translated
+// into that shape rather than merely renumbered.
+//
+// Both fixes have to replace the original event instead of adding to it, which
+// an AWTEventListener cannot do: it observes and cannot consume. Pushing an
+// EventQueue hands us each event before dispatch, so the button press is swapped
+// for the event Compose understands and the original never reaches the app.
+// Shared code stays untouched, and Windows and macOS never see this -- their
+// toolkits already number these buttons the way Compose reads them.
 //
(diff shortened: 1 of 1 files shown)


Minwalin:
I tested this PR on Fedora 44 with KDE using a touchpad. Horizontal scrolling still doesn't work on the poster carousels at the top of the Nuvio home screen. The rows don't move when using two-finger horizontal scrolling.

Everything else I tested seems to be working as expected.

ArjunKdaf:
Thanks for testing.

That top carousel doesn't scroll on any desktop platform — it's not a Linux gap. It's `HomeHeroSection` in `commonMain`: the pager is created with `userScrollEnabled = false` and driven by a custom press-and-drag gesture, so wheel and touchpad scroll events are ignored on Windows and macOS the same way. It advances on its own every 8s, or by click-dragging it, or by clicking the dots.

The poster rows below it do scroll horizontally with a touchpad here, which matches what you saw. So nothing regressed from this PR.

PR #497, opened 2026-08-27
fix(player): treat the next-episode auto play timeout as a cap

### LINKED ISSUE #454: [Bug]: Auto play timeout acts as a fixed delay floor on next-episode, not a cap
### Pre-flight checks

### App version / commit

0.1.20-alpha (MSI reports 1.1.20); also present on main@06f54b1

### Install method

### Platform

Windows (desktop)

### Desktop hardware

Ryzen 7 7800X3D + RTX 3080

### OS version

Windows 11 25H2 (build 26200)

### Area (tag)

Next Episode / Auto-play

### Bug description

The "Auto play timeout" setting is applied as an unconditional floor on the next-episode flow instead of a cap. Pressing the next-episode card always blocks for the full configured timeout, even when sources resolved seconds earlier.

In `PlayerNextEpisodeAutoPlay.kt` the outer coroutine sleeps for the whole timeout before it ever looks at the result ([L193-L203](https://github.com/NuvioMedia/NuvioDesktop/blob/06f54b16706bba4ae8cdfd46b9636a3438f29932/composeApp/src/commonMain/kotlin/com/nuvio/app/features/player/PlayerNextEpisodeAutoPlay.kt#L193-L203)):

```kotlin
if (isBoundedTimeout) {
    delay(timeoutMs)            // unconditional sleep, not raced against the result
}
applySelectionDecision(
    selectionCoordinator.onSelectionDelayElapsed(PlayerStreamsRepository.episodeStreamsState.value),
)
val completed = withTimeoutOrNull(NEXT_EPISODE_HARD_TIMEOUT_MS) { autoSelectSettled.await() }
```

`innerJob` collects `episodeStreamsState` concurrently and completes `autoSelectSettled` as soon as a stream is picked, but nothing awaits that deferred until after `delay(timeoutMs)` finishes. The selection is made on time and then sat on. The `withTimeoutOrNull(...) { autoSelectSettled.await() }` on the next line is the race that should have wrapped the delay.

With the timeout set to 15s: sources return in ~2s, the app still waits the full 15s, then runs the hardcoded 3s countdown (L211-L215). About 18s of dead time per episode.

Because `isBoundedTimeout` is `timeoutSeconds in 1..30`, the Unlimited option (`Int.MAX_VALUE`) skips the delay entirely and becomes fully event-driven. That inverts the setting: raising the timeout makes next-episode slower, and Unlimited is the fastest value. Anyone who raised it to give slow addons headroom has slowed every transition by exactly that amount.

### Steps to reproduce

1. Settings > Playback, set "Auto play timeout" to 15s.
2. Play any series episode from an addon that responds in ~1-3s.
3. Let the next-episode card appear (or press it) and time it.
4. Observe ~15s of "finding sources", then the 3s countdown.
5. Set the same setting to Unlimited and repeat. The transition now starts as soon as the addon responds.

### Expected behavior

The timeout is an upper bound. Once a stream is selected the next episode starts immediately; the timeout only matters while nothing has been selected.

### Actual behavior

Every next-episode transition takes at least the configured timeout regardless of how fast sources resolved, so Unlimited is faster than any bounded value.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

Single stream addon (self-hosted AIOStreams), no plugin scrapers, debrid handled addon-side so `resolveDebridForPlayer` is a no-op here. Measured that addon's `/stream/series/{id}.json` directly with curl, three episodes:

| episode | cold | warm |
|---|---|---|
| A | 11.59s | 1.45s |
| B | 8.19s | 2.93s |
| C | 8.79s | 0.87s |

Warm responses land in ~1-3s and are still followed by the full 15s wait, so the delay is client-side, not addon latency.

```shell

```

### Anything else? (optional)

Suggested minimal fix: race the delay against the existing deferred instead of sleeping first.

```kotlin
if (isBoundedTimeout) {
    withTimeoutOrNull(timeoutMs) { autoSelectSettled.await() }
}
applySelectionDecision(
    selectionCoordinator.onSelectionDelayElapsed(PlayerStreamsRepository.episodeStreamsState.value),
)
```

An early selection then proceeds immediately, and the timeout still bounds the wait when nothing has been chosen.

Two related things I hit while tracing this, happy to split into their own issues:

1. `NextEpisodeStreamSelectionCoordinator.decide()` gates on `state.isAnyLoading || state.groups.any { it.isLoading }`, so with several addons installed one slow addon blocks selection even when a usable stream already arrived. The desktop `OkHttpClient` uses a 60s read timeout.
2. Next-episode streams are only fetched on press (`loadEpisodeStreams`, L105). Prefetching them when the next-episode card threshold is reached would hide the addon round trip entirely. That is an architecture change, so I would want maintainer sign-off on a feature request first.

I am happy to open a PR for the timeout fix on its own if that is welcome.

(Reposting: the original report was auto-closed as unlabeled because I filed it through the API instead of the issue form.)

ArjunKdaf:
## Summary

- Race the next-episode stream selection timeout against the selection instead of sleeping through it.
- A transition now proceeds as soon as a source is picked, and the setting behaves as the ceiling it is described as.

## PR type

- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

"Stream Selection Timeout" is described as the wait for addons before selecting, but `launchPlayerNextEpisodeAutoPlay` served it as a fixed delay:

```kotlin
if (isBoundedTimeout) {
    delay(timeoutMs)
}
```

The collector started just above already settles the selection as soon as a usable stream arrives, so on a fast source the decision was made in a few hundred milliseconds and the transition then sat idle for the rest of the setting. At the default of 3 seconds that is roughly three wasted seconds on every episode, on top of the countdown that follows, and it makes the higher values on the slider unusable: choosing 20s to accommodate a slow addon costs 20s on every transition, including the ones where a source was ready immediately.

Waiting for the selection with the timeout as a bound gives the same protection for slow addons without charging fast ones for it.

## Desktop scope

`commonMain`, and required for desktop behavior: `PlayerNextEpisodeAutoPlay.kt` is the auto-play path the desktop player runs on every episode transition, and the delay is served identically on Linux, Windows and macOS. The change is one statement in that shared function, with no platform-specific branch.

## Issue or approval

Fixes #454.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

The next episode starts once a source is selected rather than once the timeout expires. Nothing changes when no source settles in time.

## Policy check

## Scope boundaries

The timeout path itself is unchanged when nothing settles: the wait still expires after `timeoutMs` and the delay-elapsed decision is applied exactly as before, followed by the same hard timeout and the same manual-selection fallback. Values outside 1..30, which never waited, are untouched.

Not changed: the three-second countdown that runs after a source is picked, the selection rules in `NextEpisodeStreamSelectionCoordinator`, the settings UI, the stored setting and its allowed values, and every other platform's player. The issue also mentions the fixed countdown; that is a separate behavior change and is deliberately not bundled here.

## Testing

Built from source on NixOS 25.11 (kernel 6.12.93, KDE Plasma on Wayland), on `Dev` at `24f63e42`.

Desktop test suite: 133 suites, 790 tests, 0 failures, 0 skipped.

Manual, on Linux desktop:

- Stream Selection Timeout set to 20s, with addons that answer quickly: the next-episode countdown started about three seconds after the episode ended, which is the countdown alone. Before this change the same setting held the transition for the full 20s first.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #454.

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/features/player/PlayerNextEpisodeAutoPlay.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/features/player/PlayerNextEpisodeAutoPlay.kt
@@ -194,8 +194,13 @@ internal fun CoroutineScope.launchPlayerNextEpisodeAutoPlay(
         val timeoutMs = timeoutSeconds * 1_000L
         val isBoundedTimeout = timeoutSeconds in 1..30
 
+        // The timeout caps how long to wait for a stream to become selectable, so
+        // it must be raced against the selection rather than served in full. The
+        // unconditional sleep made every transition cost the whole setting even
+        // when the collector above had already picked a source in the first few
+        // hundred milliseconds.
         if (isBoundedTimeout) {
-            delay(timeoutMs)
+            withTimeoutOrNull(timeoutMs) { autoSelectSettled.await() }
         }
         applySelectionDecision(
             selectionCoordinator.onSelectionDelayElapsed(PlayerStreamsRepository.episodeStreamsState.value),


KhooLy:
Please move this fix to NuvioMobile repo because its not desktop-specific.

ArjunKdaf:
Moved to NuvioMedia/NuvioMobile#1826 as you asked — the file is shared, so it belongs there.

Worth flagging while this is closed: the bug is still present in this repo too. `Dev` has since refactored this file to `NextEpisodeStreamSelectionCoordinator`, but `delay(timeoutMs)` in the bounded branch survived the refactor, so the timeout still acts as a floor here until the mobile change flows back.

PR #511, opened 2026-08-27
fix(linux): flatten jpackage classpath to work around JDK-8380085

jfb112697:
## Problem

The jpackage Linux launcher pipe-transfers the serialized application config to the JVM with a single `read`. With many jars on the classpath the serialized blob exceeds the pipe capacity, gets truncated, and the packaged launcher dies with a SIGSEGV in `jvmLauncherStartJvm` — the packaged Linux app crashes at startup.

Tracked upstream as [JDK-8380085](https://bugs.openjdk.org/browse/JDK-8380085).

## Solution

Collapse `app.classpath` in the packaged `.cfg` files to a directory wildcard (`lib/app/*`), shrinking the serialized config well under the pipe limit.

## Testing

Packaged distributable (`createDistributable`) and release AppImage build and launch correctly with the flattened classpath; playback verified end to end.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1657,6 +1657,38 @@ if (isLinuxHost) {
         }
 }
 
+if (isLinuxHost) {
+    // Workaround for JDK-8380085: the jpackage Linux launcher transfers the
+    // serialized launcher config (including every classpath entry) over a pipe
+    // with a single read/write. With many jars the buffer exceeds the pipe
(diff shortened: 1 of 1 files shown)


KhooLy:
The fix looks good, but the current Linux release pipeline also uses packageReleaseDeb, packageReleaseRpm, and packageReleaseAppImage. Please make sure the classpath workaround is applied to those package paths too, not only createReleaseDistributable.

PR #512, opened 2026-08-27
fix(linux): present player controls as a WM-managed overlay window

jfb112697:
## Problem

On XWayland the Linux player's controls were composited by snapshotting a composite-redirected WebKitGTK child window into mpv (`overlay-add`). That pipeline:

- costs a per-frame readback/upload (a 100MB-class texture push at HiDPI sizes) that made the whole UI sluggish during playback
- misbehaves driver-dependently (NVIDIA GBM failures, Mesa alpha issues), needing watchdog resets and rate limiting

## Solution

Present the controls as a separate **managed borderless toplevel**, transient for the app window (`XSetTransientForHint`) and geometry-synced to the host canvas:

- the compositor blends its ARGB surface with real alpha — native-speed controls with zero snapshot cost
- normal window management stacks/covers/minimizes it together with the app, so it no longer floats over other applications or survives minimize
- `skip_taskbar`/`skip_pager` hints plus KDE's `_KDE_NET_WM_STATE_SKIP_SWITCHER` keep it out of the taskbar, pager, and alt-tab switcher
- WebKitGTK keeps the forced software-rendering path (`NUVIO_WEBKIT_ALLOW_GPU` opts out); `NUVIO_OVERLAY_SNAPSHOT=1` restores the legacy snapshot pipeline as an escape hatch

## Testing

Validated interactively on KWin/Wayland + Mesa (amdgpu): controls render and animate at full frame rate during 4K playback, minimize/restore and alt-tab behave like any owned window, keyboard shortcuts and cursor show/hide work through player open/close cycles.

**Note:** stacked on #511 (jpackage classpath fix) — that commit is included here since this work was verified in packaged builds.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1657,6 +1657,38 @@ if (isLinuxHost) {
         }
 }
 
+if (isLinuxHost) {
+    // Workaround for JDK-8380085: the jpackage Linux launcher transfers the
+    // serialized launcher config (including every classpath entry) over a pipe
+    // with a single read/write. With many jars the buffer exceeds the pipe
+    // capacity, the read truncates, and the launcher crashes in
+    // jvmLauncherStartJvm (observed as a SIGSEGV in setenv). Collapsing the
+    // per-jar classpath entries into one directory wildcard keeps the buffer
+    // small; the JVM expands the wildcard at startup.
+    val appImageBinariesDir = layout.buildDirectory.dir("compose/binaries").get().asFile
+    tasks.matching { it.name == "createDistributable" || it.name == "createReleaseDistributable" }
+        .configureEach {
+            doLast {
+                val distributionName = if ("Release" in name) "main-release" else "main"
+                val appImageDir = File(appImageBinariesDir, "$distributionName/app")
+                appImageDir.walkTopDown()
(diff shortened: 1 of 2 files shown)


skoruppa:
yeah... no, we tried this approach. It does not work. It causes too many problems

yagomolano:
Tested this branch on a setup not covered by the original validation: hybrid GPU laptop on Hyprland.

**Setup:** Arch Linux, Hyprland (Wayland), Nuvio 0.1.21-alpha via XWayland, Intel Iris Xe + NVIDIA RTX 2050 (proprietary driver), system libmpv 0.41.

**How:** compiled `player_bridge.cpp` from `fix/linux-wm-overlay-window` against system webkit2gtk-4.1/mpv headers and swapped the resulting `libplayer_bridge.so` into the 0.1.21-alpha jar. No full rebuild needed.

**Result:** on the stock snapshot pipeline the controls overlay was completely invisible on this driver combo (video played, GUI never rendered). With this PR the controls render and are fully interactive, and the player closes back to the library correctly. Log confirms toplevel mode engaged: `overlay toplevel mode: managed transient, no reparent, no redirect`.

Happy to share build flags or fuller logs if useful.

Issue #508, opened 2026-08-27
[Bug]: unable to install 0.1.21 on macos sequoia (intel)


Liquidrag:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

Nuvio-macOS-x86_64-0.1.21-alpha.dmg

### Install method

GitHub Release desktop installer/package

### Platform

macOS (desktop)

### Desktop hardware

iMac 2015 (mid) 3.3 GHz Quad-Core Intel Core i5, AMD Radeon R9 M290 2 GB, 32 GB 1600 MHz DDR3

### OS version

macOS Sequoia 15.7.9 (24G830)

### Area (tag)

Installer / Updater

### Bug description

using in app installer and the one from github does not let me install. 

<img width="518" height="344" alt="Image" src="https://github.com/user-attachments/assets/6f41c5ae-bd42-464d-953c-f8558419f095" />

### Steps to reproduce

click update
drag nuvio icon to application

### Expected behavior

ability to drag nuvio to applications

### Actual behavior

no nuvio icon to drag over to applications folder

### Frequency

Once

### Did this work before?

Yes, it used to work

### Logs (required for crash reports)

```shell

```

PR #516, opened 2026-08-28
fix(desktop): restore macOS app icon rendering and switching

### LINKED ISSUE #515: [Bug]: Mac OS app icon not visible
### Pre-flight checks

### App version / commit

0.1.21

### Install method

### Platform

macOS (desktop)

### Desktop hardware

Mac mini m4

### OS version

26.6.2

### Area (tag)

### Bug description

1. App icon is empty when installing the app.
2. App icon not changing when i change the app icon in app settings.

### Steps to reproduce

Download 0.1.21 dmg file.
Install it.
Replace exisiting file.
App icon not visible.

### Expected behavior

Can see the app icon.
Can see the app icon if its changed from the app settings.

### Actual behavior

none

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

RakshithBhat03:
## Summary

Restore macOS app icon rendering for packaged installs and make icon changes from Settings apply to the installed app and its running Dock tile.

- Replace all seven malformed `.icns` files with valid multi-size iconsets built from the existing artwork.
- Update the actual `Nuvio.icns` file referenced by the packaged app, refresh Launch Services, and save the selection only after the update succeeds.
- Relaunch Nuvio after the old process exits so macOS reuses the pinned Dock item.
- Apply the selected icon to the running Dock tile with macOS-specific safe-area padding.
- Add desktop regression tests for bundle discovery, icon destination, relaunch ordering, and runtime Dock updates.

## PR type

- [ ] Reproducible bug fix
- [x] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

The `.icns` assets introduced in 0.1.21-alpha each declared a single `ic09` entry, which represents a 512x512 image, but embedded a 256x256 PNG. macOS rejected the malformed files, leaving a blank app icon in the DMG window, `/Applications`, and the Dock.

The Settings icon switcher also could not update the installed app. It depended on `compose.application.home`, which the packaged launcher does not set, and targeted `Contents/Resources/nuvio-app-icon.icns` even though `CFBundleIconFile` points to `Nuvio.icns`. The UI saved the new selection, but the app bundle and macOS icon remained unchanged.

This change resolves the packaged bundle from the properties its launcher actually provides, replaces the referenced icon, refreshes Launch Services, and relaunches only after the previous process exits. The restarted process also sets its selected runtime Dock image explicitly, with padding that keeps it consistent with neighboring macOS Dock icons.

## Desktop scope

macOS desktop only for the corrected package and icon-switching behavior.

The implementation is in `desktopMain`. The bundle rewrite, Launch Services refresh, relaunch helper, and Dock override run only on macOS. Windows retains its existing shortcut icon updater and `.ico` assets. Linux retains its existing window-icon behavior and `.png` package asset. No shared/common or mobile code changed.

## Issue or approval

Fixes #515

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change
- [x] UI changed only to fix a documented glitch/bug

## Policy check

## Scope boundaries

This PR only fixes the two macOS icon failures reported in #515: the blank packaged icon and Settings icon changes that did not reach the installed app.

It does not change the icon picker UI or available artwork, Windows/Linux icon behavior, mobile behavior, signing or notarization, app naming, navigation, or unrelated window management. It does not manage third-party launcher caches such as Raycast. A fresh manual DMG replacement still installs the package's default icon until Nuvio runs and restores its saved runtime selection.

The source artwork is unchanged. The 512px and 1024px entries are upscaled from the existing 256px assets because no higher-resolution source art is currently present.

## Testing

Runtime-tested on macOS 26.5.2, arm64 Mac mini.

- Ran `env JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.12/libexec/openjdk.jdk/Contents/Home ./gradlew :composeApp:desktopTest :composeApp:packageDmg -Pcompose.desktop.packaging.checkJdkVendor=false --no-configuration-cache` successfully.
- Ran `hdiutil verify composeApp/build/compose/binaries/main/dmg/Nuvio-macOS-arm64-0.1.21-alpha.dmg` successfully.
- Confirmed Apple's `iconutil` rejects all seven old assets and extracts all seven replacements into complete ten-entry iconsets.
- Confirmed the packaged `Info.plist` references `Nuvio.icns`, the packaged icon passes `iconutil`, and the runtime JAR contains the alternate `.icns` resources.
- Installed the DMG and confirmed the default icon appears in the DMG window, `/Applications`, Spotlight, and the pinned Dock item.
- Changed the icon through Settings, confirmed Nuvio exits and relaunches through the same pinned Dock item, and confirmed the selected icon updates in `/Applications` and the running Dock immediately.
- Switched between alternate icons and back to Original, confirming the Dock icon keeps the expected macOS visual size.

## Screenshots / Video

**Before:** 0.1.21-alpha DMG install window with a blank white icon:

Before: blank icon in 0.1.21-alpha DMG install window

**After:** the same DMG view with the corrected icon assets:

After: icon renders correctly with this PR's assets

**After icon switching:** the selected alternate icon appears at the expected visual size in the running pinned Dock item:

Selected alternate Nuvio icon at the expected size in the pinned macOS Dock

The Settings switching flow was also runtime-tested in the pinned Dock and `/Applications` as described above.

## Breaking changes

None.

## Linked issues

Fixes #515

Related to #508. The Intel DMG uses the same corrected package icon asset, but Intel macOS was not runtime-tested for this PR.


KhooLy:
The icon asset fixes and runtime Dock update look good, but I think there is one release blocker: MacAppIconUpdater now overwrites Contents/Resources/Nuvio.icns inside the installed .app. Our production macOS app is Developer ID signed and notarized, so modifying a sealed bundle resource after installation will invalidate the app's code signature. The current testing only uses packageDmg, so this isn't exercised against the signed/notarized release build. Could we avoid modifying the signed app bundle and keep the runtime Dock icon update separately? If persistent Finder/Application icon switching is required, it needs an approach that preserves the signature and should be verified with codesign --verify --deep --strict after switching. Also, refreshBundleRegistration() currently swallows lsregister failures, so update() can still report success.

RakshithBhat03:
Closing this because the main malformed macOS app icon used for packaging was fixed directly on Dev in ab498c93. The remaining alternate-icon and runtime-switching work needs a separate approach that preserves the signed app bundle, so it should not be merged as part of this PR. Thanks for the review.

Issue #551, opened 2026-08-30
[Bug]: Trailers stop resolving because extraction requires the rate-limited watch page and picks PO-token-gated stream URLs


aneeshtigga:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest desktop release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error.

### App version / commit

0.1.21-alpha (also reproduced on `main`, built from source)

### Install method

Built from source, and GitHub Release 0.1.21-alpha

### Desktop platform / OS

macOS 26.6 (Darwin 25.6.0), Apple Silicon (arm64)

### Bug description

In-app trailers stop resolving and show **"Unable to play trailer / No playable trailer stream found."** After a while this becomes permanent for every title until the network's IP is un-flagged by YouTube.

There are two independent causes, both verified by replaying the extractor's own request sequence with curl from the same machine.

**1. `ANDROID` / `IOS` adaptive URLs are PO-token gated.**
Their `googlevideo` URLs serve the first range and then reject every later range with 403, regardless of User-Agent (tested Chrome, Safari, `Lavf`, and each client's own UA):

| Client that minted the URL | `bytes=0-65535` | tail range | full GET |
|---|---|---|---|
| IOS (itag 137) | 206 | **403** | **403** |
| ANDROID (itag 137) | 206 | **403** | **403** |
| VISIONOS (itag 137) | 206 | 206 | 200 |

`TrailerExtractionPlatform.isUrlReachable` probes only `Range: bytes=0-0` on Android/iOS and a head+tail pair on desktop. The single head probe cannot see the gate: the first chunk always succeeds, so a poisoned URL passes validation and fails later inside the player. Desktop's head+tail probe does reject these URLs correctly, but then has nothing left to fall back to and reports `blocked stage=source reason=no_reachable_video`, which the UI shows as "No playable trailer stream found."
Only `ANDROID`'s **progressive** format (itag 18, 360p muxed) stays ungated, and nothing in the current selection logic prefers it when the adaptive ones are gated.

**2. Extraction depends on the watch page, which is the first thing YouTube rate limits.**
[...]

PR #552, opened 2026-08-30
fix(trailer): resolve trailers without the rate-limited watch page and reject PO-token-gated stream URLs

### LINKED ISSUE #584: [Bug]: Trailers stop resolving: extraction requires the rate-limited watch page and selects PO-token-gated stream URLs
### Pre-flight checks

### App version / commit

0.1.21-alpha and Dev (built from source)

### Install method

Built from source

### Platform

macOS (desktop)

### Desktop hardware

Apple Silicon (arm64)

### OS version

macOS 26.6 (Darwin 25.6.0)

### Area (tag)

Playback (start/stop/buffering)

### Bug description

Trailers fail to resolve, and once YouTube flags the network's IP nothing resolves at all. Two independent causes, both measured by replaying the extractor's own requests.

**1. `ANDROID` / `IOS` adaptive URLs are PO token gated.** They serve the first range then 403 every later range, independent of User-Agent (tested Chrome, Safari, `Lavf`, each client's own UA) and of itag:

| Client that minted the URL | bytes=0-65535 | tail range | full GET |
|---|---|---|---|
| IOS (itag 137) | 206 | 403 | 403 |
| ANDROID (itag 137/136/134/140/251) | 206 | 403 | 403 |
| ANDROID (itag 18, progressive) | 206 | n/a | 200, downloads fully |
| VISIONOS (itag 137) | 206 | 206 | 200 |

A single head-range probe cannot see this, because the first chunk always succeeds.

**2. Extraction requires the watch page, which YouTube rate limits first.** `extractPlaybackSourceInternal` fetches the watch page (1.1-1.4 MB) before every player call only to scrape `INNERTUBE_API_KEY` and `VISITOR_DATA`, and throws when it fails. Once flagged it returns 429, then a 302 to a 407-byte error page. Measured at that same moment, the player endpoint still worked:

| Video | Player API (ANDROID) | Progressive itag 18 |
|---|---|---|
| MAbjq7ydKDI | OK, with and without a key | full GET 200, 10,139,298 bytes |
| 8ZYhuvIv1pA | OK, with and without a key | full GET 200, 4,282,937 bytes |

`INNERTUBE_API_KEY` is a public constant present in every YouTube page, so the scrape is not needed.

Contributing factors: 3 player calls per extraction plus the 1.4 MB page, multiplied by hero and hover previews; stale client versions (android 20.10.35, ios 20.10.1 vs current 21.26.364 / 21.26.4); `LOGIN_REQUIRED` hits the rare clients (visionos, android_vr) long before ANDROID/IOS.

### Steps to reproduce

1. Open titles with trailers and play them, or browse Home with trailer previews enabled.
2. Trailers start failing, then stop resolving entirely.
3. With NUVIO_TRAILER_DEBUG=1 the log shows `blocked stage=extraction ... Failed to fetch watch page (429)`.
4. Confirm streams still exist: call youtubei/v1/player for the same video id with the ANDROID context and fetch the itag 18 URL. It returns 200 and downloads completely while the app reports failure.

### Expected behavior

Trailers keep resolving while YouTube serves any playable format, reducing quality if needed instead of failing.

### Actual behavior

Extraction dies on the watch page before the player API is tried, and gated URLs that pass the head-range probe then 403 during playback.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell
[TrailerDebug] extract start videoId=8ZYhuvIv1pA
[TrailerDebug] blocked stage=extraction error=IllegalStateException message=Failed to fetch watch page (429)
```

### Anything else? (optional)

A fix is implemented and tested on macOS and iOS, submitted as #552: use the public constant key and only fall back to the watch page, probe head and tail so gated URLs are rejected, prefer PO-token-free clients (visionos, android_vr) with fallback to ANDROID progressive itag 18, skip remaining clients once one resolves, refresh client versions, treat UNPLAYABLE/ERROR as final per video, and back off after a 429.

Full measurements were in #551, which the unlabeled-issue bot closed before triage; re-filing through the template so it carries the bug label. Same report for mobile: NuvioMedia/NuvioMobile#1829. Related: NuvioMedia/NuvioTV#2451.

aneeshtigga:
## Summary

Trailer resolution no longer depends on the watch page, and stream URLs that YouTube has PO token gated are now rejected before they reach the player. Resolution falls back through HLS, separate audio/video and finally `ANDROID` progressive itag 18, so trailers keep playing (at lower quality) instead of failing outright.

## PR type

## Why

Two independent problems, both measured from a real machine and described in detail in #584.

**1. Extraction required the watch page, which is the first thing YouTube rate limits.**
`extractPlaybackSourceInternal` fetched `https://www.youtube.com/watch?v=<id>&hl=en` (1.1-1.4 MB) before every player call, only to scrape `INNERTUBE_API_KEY` and `VISITOR_DATA`, and threw if it failed:

```
[TrailerDebug] extract start videoId=8ZYhuvIv1pA
[TrailerDebug] blocked stage=extraction error=IllegalStateException message=Failed to fetch watch page (429)
```

At that same moment the player endpoint still worked and the streams downloaded in full:

| Video | Player API (ANDROID) | Progressive itag 18 |
|---|---|---|
| `MAbjq7ydKDI` | `OK` with and without a key | full GET 200, 10,139,298 bytes |
| `8ZYhuvIv1pA` | `OK` with and without a key | full GET 200, 4,282,937 bytes |

`INNERTUBE_API_KEY` is a public constant present in every YouTube page (the same value this code was already scraping), so it is used directly now and the watch page is only consulted if every client returns nothing.

**2. A single head range probe cannot detect PO token gating.**
`ANDROID` and `IOS` adaptive URLs serve the first range and then reject every later range, independent of User-Agent (tested Chrome, Safari, `Lavf` and each client's own UA):

| Client that minted the URL | `bytes=0-65535` | tail range | full GET |
|---|---|---|---|
| IOS (itag 137) | 206 | 403 | 403 |
| ANDROID (itag 137, 136, 134, 140, 251) | 206 | 403 | 403 |
| ANDROID (itag 18, progressive) | 206 | n/a | 200, downloads fully |
| VISIONOS (itag 137) | 206 | 206 | 200 |

The head always succeeds, so gated URLs passed validation and then failed inside the player.

## Desktop scope

Affects macOS, Windows and Linux equally: the change is in the shared trailer extractor plus the desktop `TrailerExtractionPlatform` actual. Verified on macOS 26.6 (arm64).

`composeApp/src/androidFull` and `composeApp/src/iosFull` are touched only because `buildPlaybackSource` now takes candidate lists instead of four pre-picked candidates. Those actuals live in this repository, so they have to match or the android and ios targets stop compiling. No behavior beyond the same fix was added there.

## Issue or approval

Fixes #584.

(The first filing, #551, was closed by the unlabeled-issue bot before triage because an outside contributor cannot set labels from the API. #584 is the same report submitted through the bug template so it carries the `bug` label.)

## UI / behavior impact

- [x] No UI change

## Policy check

## What changed

- Watch page is no longer on the hot path: the public constant key is used, and `WatchConfigCache` scrapes the page only as a last resort when every client returned empty.
- Reachability requires head **and** tail ranges (the tail is only checked when `clen` is present), so gated URLs are rejected. One rejection condemns the rest of that client's adaptive formats.
- Candidate lists are passed to the platform instead of one pre-picked candidate per kind, and resolution tries HLS, separate streams, then progressive, taking the first that probes clean. A video only source is kept as a last resort so a silent trailer never wins over one with audio.
- Clients whose URLs are not gated (`visionos`, `android_vr`) are preferred, and remaining clients are skipped once one of them produced candidates. Typical extraction is now one player call instead of three.
- Client versions refreshed: android `20.10.35` to `21.26.364`, ios `20.10.1` to `21.26.4`, plus `android_vr` added. Stale versions attract bot challenges by themselves.
- `UNPLAYABLE` and `ERROR` are treated as final for that video (geo blocks, removed videos) instead of being retried.
- After an HTTP 429 the extractor backs off for 15 minutes instead of retrying per trailer, per hover and per hero. This is deliberate: request volume from one residential IP is what gets the whole network flagged (NuvioMedia/NuvioTV#2451 reports YouTube blocking every device behind the IP).
- Resolved URL cache TTL raised from 10 minutes to 3 hours, since `googlevideo` URLs carry their own `expire` of roughly 6 hours.

## Scope boundaries

Intentionally not included:

- No UI or copy changes. The existing error strings are untouched, even though "No playable trailer stream found." is currently shown both when extraction returns nothing and when the player fails to open a good URL. Happy to fix that separately if you want it.
- No persistent (on disk) cache for resolved URLs, so a restart still re-extracts. That would help further but is a larger change.
- No PO token / BotGuard support and no cookie support. Both would address the underlying bot check more completely and both deserve their own feature request.
- No changes to hero trailer or hover preview call sites.

## Testing

Manual, macOS 26.6 (Darwin 25.6.0), Apple Silicon:

1. Built unsigned release dmg from this branch (`packageReleaseDmg`), installed, and played trailers across several titles. Also built from the `0.1.21-alpha` tag to compare against the published release.
2. Reproduced the original failure first: with the previous code every trailer failed while the IP was flagged, logging `Failed to fetch watch page (429)`.
3. With this branch, trailers resolve and play on the same flagged IP. `NUVIO_TRAILER_DEBUG=1` shows the expected path: `playability client=visionos status=LOGIN_REQUIRED`, then a rejected android adaptive probe, then `selected mode=combined_fallback` on itag 18.
4. Verified the gate behaviour independently of the app by replaying the same requests with curl and comparing head range, tail range and full GET per client and per itag (tables above).
5. Checked a geo blocked trailer ("The uploader has not made this video available in your country"): it now fails on its own with `blocked stage=unplayable` and does not affect the next trailer.
6. `./gradlew :composeApp:compileKotlinDesktop` passes.

The same fix is filed for mobile as NuvioMedia/NuvioMobile#1857, where I also verified it in an iOS build.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #584. Related context: NuvioMedia/NuvioMobile#1857, NuvioMedia/NuvioTV#2451.

### DIFF
--- a/composeApp/src/androidFull/kotlin/com/nuvio/app/features/trailer/TrailerExtractionPlatform.android.kt
+++ b/composeApp/src/androidFull/kotlin/com/nuvio/app/features/trailer/TrailerExtractionPlatform.android.kt
@@ -15,6 +15,9 @@ import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
 import java.util.concurrent.TimeUnit
 
+private const val PROBE_TIMEOUT_SECONDS = 5L
+private const val PROBE_RACE_TIMEOUT_MS = 8_000L
+
 internal object TrailerExtractionPlatform {
     val defaultHeaders: Map<String, String> = mapOf(
         "accept-language" to "en-US,en;q=0.9",
@@ -34,8 +37,8 @@ internal object TrailerExtractionPlatform {
 
     private val probeClient = OkHttpClient.Builder()
         .dns(IPv4FirstDns())
-        .connectTimeout(2, TimeUnit.SECONDS)
-        .readTimeout(2, TimeUnit.SECONDS)
+        .connectTimeout(PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+        .readTimeout(PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)
         .followRedirects(true)
         .followSslRedirects(true)
         .build()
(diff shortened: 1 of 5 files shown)


KhooLy:
Try to move this PR to NuvioMobile and get core code merged in NuvioMobile first and only include desktop-specific code in here. I'll merge your implementation if your code gets merged in NuvioMobile.

PR #633, opened 2026-09-07
feat(settings): add comprehensive shortcuts screen to show all added shortcuts in one player 

AdityasahuX07:
## Summary

Added a dedicated "Shortcuts" settings page that provides a comprehensive overview of keyboard and mouse shortcuts. This includes custom mouse SVG icons, uniform Material vector arrows for directional inputs, and a navigation bypass on tablet/desktop layouts to immediately open the Shortcuts page from the sidebar rather than requiring an extra click through a mediator sub-menu.

## PR type

- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Behavior bug/regression fix
- [ ] Small maintenance only, with no UI or behavior change
- [ ] Docs accuracy fix
- [ ] Translation/localization only
- [ ] Approved larger or directional change

## Why

The application lacked a centralized way for users to view and reference keyboard and mouse shortcuts. This implements a clean UI with proper custom SVG icons, correctly sequenced shortcut preferences (e.g., M4/M5 assigned to Forward/Backward, uniform vector chevrons replacing inconsistent text symbols), and significantly improves navigation flow by directly routing to the page from the settings sidebar on desktop layouts.

## Desktop scope

Affects desktop shared code (`composeApp/src/commonMain`). Navigation routing improvements apply to desktop layouts utilizing the sidebar (`TabletSettingsScreen`), while the new Shortcuts UI page is implemented in common Compose Multiplatform code.

## Issue or approval

Fixes #634

## UI / behavior impact

- [ ] No UI change
- [x] No behavior change
- [x] UI changed only to fix a documented glitch/bug
- [ ] Behavior changed only to fix a documented bug/regression
- [ ] UI change has explicit maintainer approval
- [ ] Behavior change has explicit maintainer approval

## Policy check

- [x] I have read and understood `CONTRIBUTING.md`.
- [x] This PR is small, focused, and limited to one problem.
- [x] This PR is scoped to the desktop app, desktop packaging, desktop documentation, or shared code required for desktop behavior.
- [x] This PR is not cosmetic-only.
- [x] Any UI change fixes a linked glitch/bug and includes visual proof, or this PR has no UI change.
- [x] Any behavior change fixes a linked bug/regression or has explicit approval, or this PR has no behavior change.
- [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [x] This PR does not add dependencies, architecture changes, migrations, or product-direction changes without explicit approval.
- [x] I listed the testing performed below.

[...]

PR #639, opened 2026-09-07
feat(Shortcuts): add quick tab switching and activate search box shortcuts.

### LINKED ISSUE #625: [Bug]: The go back is not working in settings screen for esc or mouse M4 button
### Pre-flight checks

### App version / commit

0.1.21

### Install method

### Platform

Windows (desktop)

### Desktop hardware

Windows 11

### OS version

25H2

### Area (tag)

Keyboard / Mouse input

### Bug description

when press backward mouse button its not go back the previous screen of setting sub screen

Note: The esc key works.

### Steps to reproduce

1. open nuvio app
2. switch to settings
3. go deep inside the sub setting layer
4. press mouse backward

### Expected behavior

the mouse backward not go back

### Actual behavior

it should go back to previous screen same as other tabs (home, search, library)

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

### LINKED ISSUE #635: [Feature]: Add tab switching shortcuts
### Area (tag)

Keyboard / Mouse input

### Problem statement

there is no option to switch tabs directly via keyboard keys  

### Proposed solution

add 1 2 3 4 for home search library settings respectively 
also add / or 0 to open search box just like google

No, this is only a proposal

_No response_

_No response_

AdityasahuX07:
## Summary

Added global keyboard shortcuts to quickly navigate between root tabs and improved search accessibility on Desktop.

- 1, 2, 3, 4: Switches to Home, Search, Library, and Settings tabs respectively.
- / or 0: Switches to the Search tab and immediately focuses the search text box so you can begin typing.
Focus safety: All of the above global shortcuts are explicitly disabled while the search text box is actively focused, allowing users to type "1", "0", or "/" normally without triggering accidental tab switches.
Escape focus: Pressing the Escape key while the search box is focused will now seamlessly drop focus and re-enable the global navigation shortcuts.

## PR type

- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

To improve keyboard-first navigation on Desktop platforms. Users expect standard global hotkeys to switch between primary interface tabs without reaching for the mouse, and need a frictionless way to instantly jump into the search bar from anywhere in the app.

## Desktop scope

This change spans across Compose Desktop shared commonMain logic and desktopMain window event dispatching. It affects all desktop platforms (Windows, macOS, Linux). Shared native tab bridge logic was slightly extended to support explicit focus requests purely for desktop consumption.

## Issue or approval

Fixes #625, Fixes #635

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

This is strictly limited to adding the keyboard listener for the four primary root tabs and the search focus logic. Intentionally avoided adding shortcuts for deeper nested navigation, internal settings pages, or playback shortcuts to keep the PR focused and minimal. Did not apply any UI polishing or cosmetic changes.

## Testing

- Desktop OS: Windows
- Manual Flows Tested:
- Pressed 1, 2, 3, 4 (both number row and Numpad) from the Home screen; verified the app navigates successfully to Home, Search, Library, and Settings.
- Pressed /, 0, and Numpad 0 from the Library screen; verified it switches to the Search screen and brings up the text input cursor.
- Typed 1, 2, 3, 4, /, 0 directly into the focused search box; verified the characters are typed out natively and no tab routing is accidentally triggered.
- Pressed Esc while the search box was actively receiving text; verified that the cursor drops focus and the global hotkeys are re-enabled.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #625, Fixes #635

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
@@ -403,7 +403,15 @@ internal fun MainAppContent(
                 liquidGlassNativeTabBarEnabled
             ) {
                 handleRootTabClick(requestedAppTab)
+            } else if (!useNativeNavigation && isDesktop) {
+                handleRootTabClick(requestedAppTab)
             }
+        }
+    }
+
+    LaunchedEffect(Unit) {
+        NativeTabBridge.focusSearchRequests.collectLatest {
+            searchFocusRequestCount++
         }
     }
 
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
@@ -416,9 +416,6 @@ internal fun MainAppContent(
     }
 
     LaunchedEffect(Unit) {
-        // Esc and Backspace both funnel through NativeTabBridge.requestBack() (see
-        // Main.kt), so they reuse this exact same "go back" call as the mouse
-        // Back button below.
         NativeTabBridge.backRequests.collectLatest {
             navController.popBackStack()
         }
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
@@ -415,6 +415,15 @@ internal fun MainAppContent(
         }
     }
 
+    LaunchedEffect(Unit) {
+        // Esc and Backspace both funnel through NativeTabBridge.requestBack() (see
+        // Main.kt), so they reuse this exact same "go back" call as the mouse
+        // Back button below.
+        NativeTabBridge.backRequests.collectLatest {
+            navController.popBackStack()
+        }
+    }
+
     LaunchedEffect(
         nativeTabHomeTitle,
         nativeTabSearchTitle,
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/MainAppContent.kt
@@ -416,8 +416,20 @@ internal fun MainAppContent(
     }
 
     LaunchedEffect(Unit) {
+        // Esc and Backspace both funnel through NativeTabBridge.requestBack() (see
+        // Main.kt), so they reuse this exact same "go back" call as the mouse
+        // Back button below.
         NativeTabBridge.backRequests.collectLatest {
-            navController.popBackStack()
+            // Settings' sub-page navigation (e.g. Playback -> Root) is local
+            // composable state, not part of navBackStack, so popBackStack() has
+            // nothing to pop while sitting on the tab bar with Settings selected.
+            // Route through the same "step back one level" flow used when the
+            // Settings tab is reselected instead.
+            if (selectedTab == AppScreenTab.Settings && currentRoute is TabsRoute) {
+                settingsRootActionRequests.tryEmit(Unit)
+            } else {
+                navController.popBackStack()
+            }
         }
     }
 
@@ -1269,7 +1281,11 @@ internal fun MainAppContent(
                                     if (!event.changes.any { it.isConsumed }) {
                                         if (event.button == PointerButton.Back) {
                                             event.changes.forEach { it.consume() }
-                                            navController.popBackStack()
+                                            if (selectedTab == AppScreenTab.Settings && currentRoute is TabsRoute) {
+                                                settingsRootActionRequests.tryEmit(Unit)
+                                            } else {
+                                                navController.popBackStack()
+                                            }
                                         }
                                     }
                                 }
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/Components.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/Components.kt
@@ -368,7 +368,9 @@ fun NuvioInputField(
     OutlinedTextField(
         value = value,
         onValueChange = onValueChange,
-        modifier = modifier.fillMaxWidth(),
+        modifier = modifier
+            .fillMaxWidth()
+            .trackTextInputFocusForShortcutGuard(),
         singleLine = true,
         shape = RoundedCornerShape(NuvioTokens.Radius.lg),
         placeholder = {
--- a/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/NativeTabBridge.kt
+++ b/composeApp/src/commonMain/kotlin/com/nuvio/app/core/ui/NativeTabBridge.kt
@@ -36,8 +36,18 @@ internal object NativeTabBridge {
     private val _requestedTabs = MutableSharedFlow<NativeNavigationTab>(extraBufferCapacity = 1)
     val requestedTabs: SharedFlow<NativeNavigationTab> = _requestedTabs.asSharedFlow()
 
+    private val _focusSearchRequests = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
+    val focusSearchRequests: SharedFlow<Unit> = _focusSearchRequests.asSharedFlow()
+
+    var isSearchBoxFocused: Boolean = false
+
     fun requestTab(tabName: String) {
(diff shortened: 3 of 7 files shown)


KhooLy:
Two things before merge: please disable the global shortcuts while any text input is focused, not just the main Search box, and make them plain-key shortcuts so modifier combinations aren’t intercepted. Also please restore trackMaximizedBoundsForCurrentScreen() removing it is unrelated to this PR and regresses the Windows multi-monitor maximize workaround.

AdityasahuX07:
> Two things before merge: please disable the global shortcuts while any text input is focused, not just the main Search box, and make them plain-key shortcuts so modifier combinations aren’t intercepted. Also please restore trackMaximizedBoundsForCurrentScreen() removing it is unrelated to this PR and regresses the Windows multi-monitor maximize workaround.

check now

KhooLy:
The modifier fix is good, but the text-input fix still only tracks the main Search box renaming the flag to isTextInputFocused doesn’t make it global. Other text fields can still trigger the shortcuts, and now Backspace can navigate away while editing them. Also please restore and use the existing DesktopMaximizedBounds.kt helper instead of duplicating it in Main.kt; the new copy rewrites maximized bounds on every move/resize, which the existing implementation intentionally avoids. I’d also remove the new global Backspace binding since #625 only requires the mouse back fix and already says Esc works.

AdityasahuX07:
> The modifier fix is good, but the text-input fix still only tracks the main Search box renaming the flag to isTextInputFocused doesn’t make it global. Other text fields can still trigger the shortcuts, and now Backspace can navigate away while editing them. Also please restore and use the existing DesktopMaximizedBounds.kt helper instead of duplicating it in Main.kt; the new copy rewrites maximized bounds on every move/resize, which the existing implementation intentionally avoids. I’d also remove the new global Backspace binding since #625 only requires the mouse back fix and already says Esc works.

The backspace is intentionally add it was my original idea of shortcut feature because the backspace is easily accessible specially for the right handed users.

KhooLy:
Yeah no app using that. We dont need Backspace

AdityasahuX07:
> Yeah no app using that. We don't need Backspace

in future when the arrow keys will implemented the Backspace is essential

AdityasahuX07:
> Yeah no app using that. We dont need Backspace

Stremio, all web app and browser uses the Backspace to go back universally

KhooLy:
Arrow-key navigation doesn’t make Backspace essential. Esc already covers keyboard back navigation without conflicting with text editing, so I’d prefer to keep Backspace unbound globally. Also no, there is no "backspace" shortcut universally.

KhooLy:
When we add bindable shortcuts, Users can do whatever they want with shortcuts. Currently we only set universal shortcuts. Which backspace is used for deleting text. Esc is global and universal for going back.

AdityasahuX07:
> Arrow-key navigation doesn’t make Backspace essential. Esc already covers keyboard back navigation without conflicting with text editing, so I’d prefer to keep Backspace unbound globally. Also no, there is no "backspace" shortcut universally.

backspace just add extra option if u dont want to use then just dont use this nothing changes for you but adds extra option for the user who used to with backspace as go back is this is the option for that, so in conclusion nothing removes just add extra option that not affect other user they can chose their own comfort.

personally, I am always hit to backspace every time then press the esc because backspace is not working yet.

since majority people come from stremio or use side by side with nuvio in this case backspace is very important feature.

KhooLy:
I get the preference argument, but a global Backspace binding is not just an extra option unless it’s configurable. Backspace already has an editing function, so I don’t want it bound globally by default. If we add configurable shortcuts later, users can opt into Backspace then.

AdityasahuX07:
> When we add bindable shortcuts, Users can do whatever they want with shortcuts. Currently we only set universal shortcuts. Which backspace is used for deleting text. Esc is global and universal for going back.

one more things to add that I personally in my workflow to use like this - esc is used to close video player and backspace is used to go back. This is let me help to not close the video player accidently esc for me works as kill switch of video player.

this is reason I added 2 go back keys

as already said it does not disable esc when backspace is added both work same to go back

KhooLy:
I understand your workflow, but I don’t want Backspace as a default global binding. Let’s keep Esc for now and revisit Backspace when shortcuts become configurable.

AdityasahuX07:
> I understand your workflow, but I don’t want Backspace as a default global binding. Let’s keep Esc for now and revisit Backspace when shortcuts become configurable.

please I insist keep this since this not affect anyone who use esc but who migrated or use stermio side by side this is life saver

KhooLy:
I understand, but I’m not accepting Backspace as a default global binding. Please remove it from this PR. We can revisit it later as a configurable shortcut.

AdityasahuX07:
> I understand, but I’m not accepting Backspace as a default global binding. Please remove it from this PR. We can revisit it later as a configurable shortcut.

Ok we will do this in configurable shortcut

AdityasahuX07:
> The modifier fix is good, but the text-input fix still only tracks the main Search box renaming the flag to isTextInputFocused doesn’t make it global. Other text fields can still trigger the shortcuts, and now Backspace can navigate away while editing them. Also please restore and use the existing DesktopMaximizedBounds.kt helper instead of duplicating it in Main.kt; the new copy rewrites maximized bounds on every move/resize, which the existing implementation intentionally avoids. I’d also remove the new global Backspace binding since #625 only requires the mouse back fix and already says Esc works.

Done, check this now

KhooLy:
Thanks for the updates. The modifier handling and maximized-bounds helper look fixed, and Backspace has been removed.  
The text-input issue is still unresolved, though. Main.kt only checks NativeTabBridge.isSearchBoxFocused, and the flag is only updated by SearchScreen.kt. Other text fields, such as settings, addon URL, or server URL inputs, can still trigger the global 1–4, /, 0, and Esc shortcuts while typing.  
Please make the focus guard global for any focused text input. Once that is fixed, I’ll be happy to merge #633 and #639 together.

AdityasahuX07:
> Thanks for the updates. The modifier handling and maximized-bounds helper look fixed, and Backspace has been removed. The text-input issue is still unresolved, though. Main.kt only checks NativeTabBridge.isSearchBoxFocused, and the flag is only updated by SearchScreen.kt. Other text fields, such as settings, addon URL, or server URL inputs, can still trigger the global 1–4, /, 0, and Esc shortcuts while typing. Please make the focus guard global for any focused text input. Once that is fixed, I’ll be happy to merge #633 and #639 together.

Done. 

- the shortcut input 1 2 3 4 / 0 not trigger while typing in search box.
- esc to revoke the focus form search box.

- the esc is intentionally allowed to revoke the search box so we can easily switch the tabs.
- The reason if esc I will disable then require extra mouse use to revoke the search box focus this totally break the main motive of using keyboard if u need the mouse if u touch the mouse then why u use the 1 2 3 4 tab switching.
- so the esc is important to be active input even the search box is active. The 1 2 3 4 0 / is disabled only because it can be use in the movie name so it triggers the unwanted tab switching, but esc is intentional command which must active no matter what we are doing.

See the test build screen rec. -

everything working as expected.

https://github.com/user-attachments/assets/b7604e1c-8654-422b-bd68-0ba0e2cdace0

KhooLy:
The Backspace, modifier and maximized-bounds issues are fixed. The remaining text-input issue still isn’t fixed globally, though. trackTextInputFocusForShortcutGuard() currently only gets attached to the main Search field. Library search, Settings search, API/secret fields, addon/server URL fields and other text inputs still never register focus, so 1–4, /, 0 and Esc can still be intercepted while typing in them. Please apply the guard to all text-input paths (preferably through shared text-field components where possible), then rebase onto current Dev. Also clean up the stale Backspace comment in MainAppContent.

AdityasahuX07:
> The Backspace, modifier and maximized-bounds issues are fixed. The remaining text-input issue still isn’t fixed globally, though. trackTextInputFocusForShortcutGuard() currently only gets attached to the main Search field. Library search, Settings search, API/secret fields, addon/server URL fields and other text inputs still never register focus, so 1–4, /, 0 and Esc can still be intercepted while typing in them. Please apply the guard to all text-input paths (preferably through shared text-field components where possible), then rebase onto current Dev. Also clean up the stale Backspace comment in MainAppContent.

Done.

AdityasahuX07:
Closing merge your implementation

PR #346, opened 2026-08-12
feat(desktop): add native Picture-in-Picture support

GuickerZ:
## Summary
Adds native desktop Picture-in-Picture support as an independent, borderless, always-on-top video window while preserving the existing player handle and playback session.

## PR type
- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Behavior bug/regression fix
- [ ] Small maintenance only, with no UI or behavior change
- [ ] Docs accuracy fix
- [ ] Translation/localization only
- [ ] Approved larger or directional change

## Why
The desktop player did not previously provide a native PiP mode. This adds the missing capability with a browser-like floating video window, a dedicated compact control layout, restore paths, native resize/drag support, dynamic localization, and stable focus/handle ownership.

## Desktop scope
Windows and macOS desktop player surfaces. The small common/platform expect/actual changes only provide the shared compile-time contract required to expose the desktop behavior; they do not enable PiP on Android or iOS.

## Issue or approval
No linked issue: requested and approved as a desktop feature by the repository owner.

## UI / behavior impact
- [ ] No UI change
- [ ] No behavior change
- [ ] UI changed only to fix a documented glitch/bug
- [ ] Behavior changed only to fix a linked bug/regression
- [ ] UI change has explicit maintainer approval
- [x] Behavior change has explicit maintainer approval

## Policy check
- [x] I have read and understood CONTRIBUTING.md.
- [x] This PR is small, focused, and limited to one problem.
- [x] This PR is scoped to the desktop app, desktop packaging, desktop documentation, or shared code required for desktop behavior.
- [x] This PR is not cosmetic-only.
- [x] Any UI change fixes a documented glitch/bug and includes visual proof, or this PR has no UI change.
- [x] Any behavior change fixes a linked bug/regression or has explicit approval, or this PR has no UI change.
- [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [x] This PR does not add dependencies, architecture changes, migrations, or product-direction changes without explicit approval.
- [x] I listed the testing performed below.

## Scope boundaries
No Android/iOS PiP behavior, no backend changes, no navigation changes, and no playback-engine replacement. Global hardware media-key handling remains isolated in PR #344.

## Testing
- :composeApp:buildWindowsPlayerBridge
- :composeApp:compileKotlinDesktop
- :composeApp:desktopTest
[...]

PR #642, opened 2026-09-09
feat(desktop): add native Picture-in-Picture support

### LINKED ISSUE #125: [Feature]: Picture in Picture pop out video during playback
### Area (tag)

Playback

### Problem statement

Add a Picture-In-Picture button on the desktop app, not everyone has multiple monitors and sometimes we like to view media while doing other things on the computer. A simple solution is a button to pop out the video player, keeping it on top of other apps allowing users to resize the window and drag it anywhere on the screen. Additional features on the PiP window could be Play/Pause button (mandatory) and optional buttons for ffwd and going back in increments of 5/10 seconds, as well as adjust the volume when hovering the popped out video tile. 

PiP is an essential feature that is pretty much universal on media playback apps and websites (usually in-built into the browser)
Proposed solution

### Proposed solution

Offer a button to pop out the video playback in a separate window above all the other applications on the screen

No, this is only a proposal

_No response_

_No response_

mrnasil:
## Summary
Adds native desktop Picture-in-Picture as an independent, borderless, always-on-top video window while keeping the existing player handle and playback session. This is a rebase of #346 onto current Dev with follow-up fixes: leaked Windows media-key handlers removed, Linux JNI stubs added, Escape restores PiP instead of closing the player, and the PiP control is hidden on Linux.

## PR type
- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why
The desktop player did not provide a native PiP mode. Users without multiple monitors need a floating always-on-top video window with play/pause, seek, volume, resize, and drag. This implements the desktop feature requested in #125.

## Desktop scope
Windows and macOS desktop player surfaces. Shared expect/actual changes only provide the compile-time contract; they do not enable PiP on Android, iOS, or Linux.

## Issue or approval
Closes #125. Requested as a desktop Picture-in-Picture feature. Original implementation in #346.

## UI / behavior impact
- [ ] No UI change
- [ ] No behavior change
- [x] UI change has explicit maintainer approval

## Policy check

## Scope boundaries
No Android/iOS/Linux PiP behavior, no backend changes, no navigation changes, and no playback-engine replacement. Global hardware media-key handling from closed PR #344 is not included.

## Testing
- Rebased onto current Dev with no remaining merge conflicts
- Linux JNI exports added for beginWindowDrag, setWindowResizable, and reparentSurfaceNative so the desktop shared bridge still loads
- Manual review of Windows PiP toggle, restore, Escape, P shortcut, compact controls, resize, and drag paths in the native controls overlay

## Screenshots / Video
This is a new desktop UI/behavior feature. Screenshots from the original #346 implementation:
image
image
image
pip-window

## Breaking changes
None.

## Linked issues
Closes https://github.com/NuvioMedia/NuvioDesktop/issues/125


GuickerZ:
Looks great, thanks for the follow-up @mrnasil! Tested locally on Windows and working as expected.

GuickerZ:
I went over the PiP implementation to fix a few things I had missed in my initial PR, and opened a PR on your fork so you can pull the changes with one click:
https://github.com/mrnasil/NuvioDesktop/pull/1

What was fixed/added:
- Fixed window dragging (it was snapping and getting stuck previously). Now it drags smoothly anywhere on the background.
- Fixed window resize so it properly follows the video aspect ratio.
- Set the PiP window/taskbar icon to match the user's selected Nuvio app icon instead of the default Java one.
- Fixed button alignment on the PiP header so there is no awkward trailing gap.
PiP lock button
PiP locked overlay


- Kept the "You're watching" screen in PiP when paused:

PiP watching screen


- Cleaned up the diff (removed unused strings/params and reverted mobile player files I had touched by mistake).

Tested everything locally on Windows and it's working smoothly.

mrnasil:
Pulled your follow-up from https://github.com/mrnasil/NuvioDesktop/pull/1 and kept the lock control.

One behavior change on top: a locked PiP window can still be dragged. Unlock is only from the padlock badge, so accidental play/pause/seek stays blocked without trapping the always-on-top window in place. Lock/unlock labels now use the existing localized player strings.

ThEgIGN:
Hello, just had a chance to test this since it got merged. Sorry for the wall of text about to come.

Had a problem with `DWMWA_BORDER_COLOR` and `DWMWA_CAPTION_COLOR` variables, but fixed it. While fixing it, I noticed those variables are Windows 11 only (I might be wrong). So, maybe include something like this for people on Windows 10 and lower in windows `player_bridge.cpp`:

```
#ifndef DWMWA_BORDER_COLOR
#define DWMWA_BORDER_COLOR 34
#endif

#ifndef DWMWA_CAPTION_COLOR
#define DWMWA_CAPTION_COLOR 35
#endif
```

Next episode button unfortunately sneaked in. Looking at the comment in `controls.css`, it wasn't supposed to. And that would explain why it completely breaks both PiP and Nuvio itself if you press that button, since it's not supposed to be there.
Simply adding `.player.pip-mode [data-command="playNextEpisode"],` to an already existing `display: none !important` fixes it.

Mouse clicks (both regular and double) are ignored. I was going to say I don't agree with this design choice, but after seeing double click has an action to close PiP and that isn't working, I realized it was a bug.
You can't pause/play while pressing the left click, but most importantly, you can't close the subtitle/audio menu with a mouse click, you have to use the Escape key.
The problem is `dragWindow` in the `pointerdown` event. `Pointerdown` event triggers first and `click` and `dblclick` events will never trigger after.
I tried to fix it by using the `pointermove` event listener (similiar approach to speed boost getting canceled before it starts) and only start dragging if you hold your mouse and move the pointer by more than 12 units. Problem with that solution lies in `ReleaseCapture()` from `player_bridge.cpp`. Because of this, `pointerup` and `pointercancel` events will never trigger and thus won't reset variables that check if the pointer is down and that dragging has started.

The Nuvio window that remains after PiP is enabled has Mobile player gesture controls enabled. You can hold to activate speed boost, you can double tap on left/right to seek and you can hold and drag left/right to precisely seek. Also, the pause overlay shows over the "Restore to main Player" button (Tap completely changed pause overlay for mobile recently, so that might be why).

The Nuvio window that remains after PiP minimizes and maximizes slower, sometimes taking 2–3 seconds.

While resizing, the window glitches out. I understand why it's glitching out, since it's fighting me in order to respect aspect ratio and not show unnecessary black bars, but it's a bit harsh on the eyes.
Not sure if possible, but it'd be amazing if you could change that behavior to follow, for example, YouTube PiP. No matter which side of the window you try to resize it from, it will always start resizing as if you selected a corner. That way, you can always respect the aspect ratio of the video and, hopefully, stop flickering. :D

Anyway, great job with this. This is really nicely done. These are just a couple of small issues I found, some of them critical and some that can be easily fixed at a later date.

GuickerZ:
I'll fix it later, ok?

Issue #565, opened 2026-08-31
[Feature]: Support MPRIS


Celinhovp:
### Area (tag)

Platform-specific (Windows / macOS / Linux)

### Problem statement

No support MPRIS control.

### Proposed solution

With MPRIS support, Nuvio would be able to:

Display the current song, artist, album, and album artwork in desktop media widgets.
Work with GNOME extensions such as Dynamic Pill and Media Controls.
Integrate with KDE Plasma, Waybar, playerctl, and other Linux desktop environments.
Support Play/Pause, Next, Previous, Seek, and media key controls.

Many Linux users rely on MPRIS for a consistent media experience, and adding this feature would make Nuvio feel much more integrated with the desktop.

Thank you for considering this feature!

### Are you planning to implement this yourself?

No, this is only a proposal

PR #653, opened 2026-09-10
feat(linux): add MPRIS media controls

wiktorekdev:
## Summary

Expose the Linux player through MPRIS so desktop media controls and tools such as `playerctl` can read playback state and control playback, seeking, volume, speed, and episode navigation.

## PR type

- [ ] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

This is a new feature associated with #565 and does not yet have explicit maintainer approval.

## Why

Nuvio currently does not expose its player through the standard Linux media-control interface requested in #565.

## Desktop scope

Linux desktop only. The D-Bus implementation and dependencies are included only in Linux builds. Other desktop platforms use a no-op bridge.

## Issue or approval

Related to #565. The request is open but does not contain explicit maintainer approval.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

No UI, URI-handler, taskbar, playback-setting, repeat, or shuffle changes are included.

## Testing

- `./gradlew :composeApp:desktopTest --tests com.nuvio.app.features.player.desktop.MprisBridgeTest --no-daemon`
- Live session D-Bus test with a local video
- `playerctl` status, pause, play, and volume checks
- Verified separate root/player property maps and exported player metadata
- `git diff --check`

The focused desktop test suite passed on the current upstream `Dev`. KDE widgets, lock-screen controls, and Bluetooth media buttons were not tested.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Related to #565.


KhooLy:
Looks promising, but I found three blockers: MprisBridgeTest is in generic desktopTest while both dbus-java and MprisObject only exist on Linux, so non-Linux desktop tests will not compile. Previous dispatches playPreviousEpisode, but there is no handler for that event, so previous-episode navigation is currently broken. Also, the MPRIS track ID is regenerated whenever any PlayerNowPlayingInfo field changes; it should remain stable for the same media item and only change when the actual track/item changes.

wiktorekdev:
I’ll take a look

wiktorekdev:
Fixed all three blockers:

- Moved the MPRIS bridge tests to the Linux-only desktop test source set.
- Wired the previous-episode event into the existing episode playback flow.
- Made MPRIS track IDs stable for the same media item and rotate only when the item changes.

While testing the real app, I also found that metadata published before the native player attached was being cleared during MPRIS registration. That lifecycle case is fixed and covered by regression tests.

Verified with the full desktop test suite: 950 tests passing. I also tested the flow in Nuvio—sending Previous through MPRIS while paused near the beginning of an episode correctly opened the previous episode.

KhooLy:
The previous three blockers are fixed, thanks. I found one remaining logic issue: CanGoPrevious is derived from the raw episodeItems index, while the actual Previous action resolves episodes after sorting by season/episode. With unsorted addon episode lists, MPRIS can advertise the wrong capability. Please derive the capability from the same episode-ordering logic used by resolvePreviousEpisode(). Also, the new generic previous-episode/autoplay changes live in commonMain while Mobile still has the old shared implementation; please upstream that shared behavior to Mobile first or keep the MPRIS change isolated to the desktop seam.

wiktorekdev:
Pushed the follow-up fix.

* `CanGoPrevious` now uses the same season/episode ordering as the actual Previous action, so unsorted addon episode lists no longer produce a mismatched capability.
* The previous-episode ordering/resolution is kept desktop-only.
* I removed the earlier shared Previous/autoplay changes from `commonMain`; Mobile behavior is unchanged.
* The desktop Previous action now passes the resolved original episode index through a small platform-neutral `playEpisodeAtIndex` seam, reusing the existing autoplay pipeline without duplicating it.

I also re-tested MPRIS live: Previous, Next, Play/Pause, Seek, Volume, Rate and Stop are working, including Previous from S1E2 -> S1E1.

KhooLy:
Thanks, the previous fixes look good. The `CanGoPrevious` ordering issue and the Linux-only test setup are resolved.

I still see two blockers before merging:

1. The stable MPRIS track ID fix appears to have been regressed in the latest commit. `itemId` was removed again, and `MprisObject.update()` now rotates the track ID whenever `PlayerNowPlayingInfo` changes. Metadata changes for the same media item should not generate a new MPRIS track ID. Please restore a stable media identity and add a regression test that verifies metadata-only updates keep the same track ID.

2. `playEpisodeAtIndex` and its event handler are still added in `commonMain`. Since NuvioMobile is the canonical upstream for shared player behavior, please either upstream this shared seam to Mobile first or keep the MPRIS previous-episode path entirely desktop-specific.

After those are addressed, please update/rebase the branch onto the current `Dev` and rerun the desktop/Linux tests.

wiktorekdev:
not ready yet

wiktorekdev:
I opened NuvioMedia/NuvioMobile#1928 to move playEpisodeAtIndex upstream.

Once that’s merged, I’ll rebase this PR, use the shared action for MPRIS Previous.

The stable track ID fix is already back in place with a regression test.

PR #680, opened 2026-09-15
feat(desktop): macOS full-window-content title bar

azooz5959:
## Summary

Adopts Apple's native full-window-content title bar on macOS, so the traffic
lights sit directly over the app's own artwork instead of on a separate grey
title bar strip. Also realigns the streams back button so it matches the meta
details screen and clears the traffic lights.

macOS only — Windows and Linux code paths are untouched.

### Before / after

| | |
|---|---|
| **Before** | A grey native title bar strip with the "Nuvio" title, with a visible seam above the content. |
| **After** | Content extends under the traffic lights; no strip, no seam. |

> Screenshots attached in a comment below.

## Changes

**`DesktopWindowChrome.kt`** — new `configureMacosWindowBeforePeer(window)` sets
three root pane client properties:

```kotlin
putClientProperty("apple.awt.fullWindowContent", true)
putClientProperty("apple.awt.transparentTitleBar", true)
putClientProperty("apple.awt.windowTitleVisible", false)
```

**`Main.kt`** — switched `Window(...)` to `SwingWindow(..., init = ::configureMacosWindowBeforePeer)`.

This matters: AWT reads those client properties when it creates the native
`NSWindow` peer and ignores them afterwards, so they have to be set *before*
the window is displayable. `SwingWindow`'s `init` hook is the point where that
is guaranteed. A `check(!window.isDisplayable)` assertion in
`configureMacosWindowBeforePeer` fails loudly if that invariant is ever broken
by a future refactor.

No `NSWindow.styleMask` mutation and no `NSVisualEffectView` insertion into the
AWT content view — both interfere with the Skiko Metal layer.

**`StreamsScreen.kt`** — under full-window-content, `WindowInsets.safeDrawing`
reports no top inset on macOS, so the back button landed on top of the traffic
lights. The desktop branch now uses the same geometry as the meta details
screen (32dp top, standard page horizontal inset, 48dp button, 24dp icon,
translucent black container). Mobile metrics are unchanged.

## Test plan

- [x] Built and launched on macOS 15 (arm64, Oracle JDK 26, Metal backend), both before and after the change.
- [x] No `SIGSEGV` / no `hs_err_pid*.log` produced across repeated launches.
- [x] Window dragging by the title bar area, minimise, zoom and fullscreen all still work (native controls unchanged).
- [x] No content shift or clipping at the top of any screen.
- [x] Streams back button no longer overlaps the traffic lights, and now sits in the same place as on the details screen.
- [ ] Not verified on Windows/Linux — those paths are unmodified by this PR.


azooz5959:
titlebar_before_after

PR #707, opened 2026-09-18
fix(macos): make volume changes audible immediately on the avfoundation AO

### LINKED ISSUE #708: [Bug]: macOS volume changes take seconds to become audible (avfoundation audio output)
### Pre-flight checks

### App version / commit

0.1.24-alpha

### Install method

### Platform

macOS (desktop)

### Desktop hardware

MacBook Air (M1, 2020), 16 GB

### OS version

macOS 26.6.2 (25G83)

### Area (tag)

### Bug description

While a stream is playing, any volume change from the player controls is applied to the on-screen slider immediately but is only *heard* about 2 to 4 seconds later. This affects every way of changing volume: scrolling over the volume control, clicking a point on the slider, dragging it, the keyboard volume shortcuts, and mute (the mute button or the `M` key keeps playing sound for a couple of seconds before going silent).

The player software path is not the bottleneck. With timing instrumentation in the native bridge, the interval from the controls page posting `volumeChange` to mpv reporting the new `volume` value was 1 to 4 ms on every command. The delay is in the audio output.

On macOS the player uses mpv's `avfoundation` audio output (`ao=avfoundation,coreaudio`, introduced by d9d0c423 for spatial audio). mpv's `volume` property is software gain applied *before* the audio output's buffer, and this output buffers deeply. mpv logs on startup:

```
[ao/avfoundation] device buffer: 96000 samples
[ao/avfoundation] using soft-buffer of 96000 samples
```

At 48 kHz that is roughly 2 s + 2 s of audio already queued at the old gain, so a softvol change only becomes audible once that queue has drained. The output's own renderer volume (mpv property `ao-volume`) is applied at the output stage and takes effect immediately, which is what the fix uses.

### Steps to reproduce

1. Open the desktop app on macOS and start playing any movie or episode from any source.
2. Once audio is playing, hover the volume control and scroll the wheel a few notches down, or click a clearly lower point on the volume slider.
3. Watch the slider and listen.
4. Press `M` (or click the mute button) and listen.

### Expected behavior

The audible volume follows the slider immediately (within a frame or two), and mute silences playback immediately, as it does on the other desktop platforms and in other macOS players.

### Actual behavior

The slider moves at once, but the sound stays at the previous level for about 2 to 4 seconds and then jumps to the new level. Mute keeps playing sound for the same period before going silent. No error text is shown; the app does not freeze.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

**Media details (optional):**

- Content type: any (reproduced with series episodes and movies)
- Stream/source: any; independent of the addon or source used
- Audio: observed with 48 kHz stereo tracks; the delay scales with the sample rate and buffer size, not the content

**Logs (required for crash reports):**

Not a crash. Relevant mpv log lines from the app at playback start (unified log, process `Nuvio`):

```
[ao/avfoundation] device buffer: 96000 samples
[ao/avfoundation] using soft-buffer of 96000 samples
```

Timing instrumentation over 44 consecutive volume commands during a scroll gesture (instrumented build, not shipped): controls page `volumeChange` to mpv `volume` property applied = 1 to 4 ms, mean 2.9 ms. The audible change followed 2 to 4 s later.

```shell

```

### Anything else? (optional)

- Workaround: none from the UI. Changing the Mac's system volume is immediate, which points at the app's output buffering rather than the device.
- Not a duplicate of #563 (Windows, clicks ignored) or #618 / #687 (boost above 100% via wheel/keyboard). #657 (macOS, playback opens muted) is a different symptom but the same output stage, so it may be worth re-checking with the fix below since it sets the renderer volume explicitly when the output comes up.
- Fix: route the 0 to 100% part of the level through `ao-volume` on the avfoundation output, keep mpv's softvol at unity, and use softvol only for boost above 100%. Gated on `current-ao == avfoundation` because on the coreaudio fallback `ao-volume` is the system device volume. PR: NuvioMedia/NuvioDesktop#707.

tavyandy97:
## Summary

On macOS, volume changes from the player controls (scroll, slider click/drag, keyboard, mute) took about 2 to 4 seconds to become audible even though mpv applied them within milliseconds. The macOS `avfoundation` audio output buffers roughly 4 s of audio (96000-sample device buffer plus 96000-sample soft buffer at 48 kHz) and mpv's software volume is applied before that buffer.

This change routes the 0..100% part of the volume through the output's own renderer volume (mpv's `ao-volume`, applied at the output and effective immediately) and keeps mpv's softvol at unity; softvol is used only for boost above 100%. The effective level is read back as the product of both so the controls keep showing the level the user set. The split is gated on mpv reporting `current-ao == avfoundation`, observed through a small wakeup-driven mpv event drain, because on the coreaudio fallback `ao-volume` is the system device volume.

One file changed: `composeApp/src/desktopMain/native/macos/player_bridge.mm`.

## PR type

- [x] Reproducible bug fix
- [ ] UI glitch/bug fix
- [ ] Docs accuracy fix

## Why

- Old behavior: every volume command wrote mpv's `volume` (softvol) and the change was heard once the avfoundation output had drained the audio already queued at the old gain, about 2 to 4 s later. Mute kept playing sound for the same period.
- Broken behavior: the slider moves immediately while the sound lags by seconds, on every volume path. Measured on an instrumented build over 44 commands: control page to mpv `volume` applied with a median of 1.1 ms (max 60.5 ms); audible change 2 to 4 s later. Root cause is the output buffer, not the software path.
- New behavior: levels up to 100% are applied through `ao-volume` and are audible immediately; boost above 100% still rides softvol, so only the boost part keeps the buffer delay. Any other audio output keeps the previous softvol-only behavior.
- Tested manually on the installed .app (see Testing).

## Desktop scope

macOS only. The change is confined to the macOS native player bridge (`player_bridge.mm`). No shared/common code, no Kotlin, no controls page, no Windows or Linux bridge changes. Windows and Linux use different mpv audio outputs and are not affected by this buffer.

## Issue or approval

Fixes #ISSUE_NUMBER

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Intentionally not changed:

- Windows and Linux bridges, the Kotlin controller and the controls page: untouched.
- The existing 500 ms property poll that reports playback state to the controls page: unchanged; it now reads the effective level (softvol times `ao-volume`).
- Boost above 100% still goes through softvol and keeps a short delay for the boost portion only. Making boost instant would need a different mechanism and is out of scope.
- The `adjustVolume` JNI entry point (no Kotlin caller today) is routed through the same write path for consistency, nothing else.
- No UI polish, no behavior tweaks beyond the fix, no refactors.

## Testing

Environment: macOS 26.6.2 (25G83), MacBook Air (M1, 2020), Apple silicon build.

- Built the native bridge with `./gradlew :composeApp:buildMacosPlayerBridge` and the release app with `./gradlew :composeApp:createReleaseDistributable`, installed the resulting `Nuvio.app` and played streams from several sources.
- Manual checks on the installed app: volume by scroll wheel over the control, by clicking a point on the slider, by dragging, by keyboard shortcuts, mute via the button and the `M` key. The audible level now follows immediately in every case, and the slider holds the chosen level after the next state poll (previously it also held; verified it still does with the split).
- Boost above 100% still works; the boost part lags as described in Scope boundaries.
- Verified with an instrumented build (not part of this PR) that mpv reports `current-ao=avfoundation`, that the `ao-volume` and `volume` writes both return success, and that the observed property values match the requested split.
- Unit tests: `./gradlew :composeApp:desktopTest` passes on this branch (Kotlin side is unchanged by this PR).
- Not tested: the coreaudio fallback path (requires avfoundation to fail to initialize). It keeps the previous softvol-only code path.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes https://github.com/NuvioMedia/NuvioDesktop/issues/708

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tavyandy97:
@KhooLy When you get a chance, could you please review this?

